PR #2, opened 2016-05-20
Do not hardcode marathon to localhost

m0ppers:
I am trying to setup the adminrouter locally so I can develop without having to install a real DC/OS cluster somewhere but can stay on my machine. For that I have a some docker containers which quickly spawn a local mesos cluster. On top of that I am currently creating a docker container containing the adminrouter. This works fine.

However the marathon URL is hardcoded to 127.0.0.1 instead of facilitating mesos service discovery. Not sure if that was intentiional but without this change it won't work for me as my marathon is running somewhere else :S


alberts:
For local dev, you might give https://github.com/dcos/dcos-docker a try. Right now AdminRouter's contract is to specifically talk to its local Marathon.

PR #3, opened 2016-05-25
Modify /slave rule to accept /agent too.

kaysoky:
The slave->agent rename (in Mesos 0.29+) affects a couple of the endpoints.
For backwards compatibility, this rule should take both old and new terms.


alberts:
Fixed in another PR.

PR #9, opened 2016-06-28
Adding agent-nginx for Open DC/OS

malnick:

PR #10, opened 2016-07-08
Add /pkgpanda/ location

branden:
This adds a location for the Pkgpanda API on all cluster nodes.


branden:
This PR depends on the agent adminrouter from #9.

PR #11, opened 2016-07-12
remove nonexistent networking API endpoint

alberts:


alberts:
@sargun I don't think we want to litter the Open repos with those details.

sargun:
@alberts then 👍

PR #13, opened 2016-07-15
Updated maximum client body size

titosand:
Updated the maximum client body size configuration value to solve a 413 Request Entity Too Large Error in Nginx when uploading a plugin > 1MB in Jenkins.


ssk2:
@spahl, @alberts: will we be able to get this PR into 1.8 EA? It's necessary for several of our Developer Agility services that will be coming to the Universe soon :)

PR #14, opened 2016-07-19
fixed port enforcement

bernadinm:
This allows others nginx to proxy the $http_host so that other proxies can call adminrouter without issue.

Tested this on my end and it works for both the existing adminrouter and haproxy

### DIFF
--- a/nginx.master.conf
+++ b/nginx.master.conf
@@ -150,7 +150,7 @@ http {
 
         location ~ ^/service/(?<serviceid>[0-9a-zA-Z-.]+)$ {
             # Append slash and 301-redirect.
-            rewrite ^/service/(.*)$ /service/$1/ permanent;
+            rewrite ^/service/(.*)$ $scheme://$http_host/service/$1/ permanent;
         }
 
         location ~ ^/service/(?<serviceid>[0-9a-zA-Z-.]+)/(?<url>.*) {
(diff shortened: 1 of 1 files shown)


jgehrcke:
I'd love to get a better understanding of the issue that this patch is supposed to solve. Do we have a Jira ticket for that?

We need to be very careful with changes in the /service route, and thoroughly test them to make sure that we do not break existing setups. We do not have automated tests that ensure preventing breakage.

By default, nginx' `proxy_redirect` changes the Location header of upstream responses so that the ["primary server’s name and port"](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect) are inserted. That's the current behavior, as the `replacement` string contains neither scheme, host, port. If Admin Router is the first reverse proxy in the chain (from the user agent's point of view), this behavior makes sense.

I infer from your description that the problematic scenario is when there is (at least) one additional reverse proxy _in front of_ Admin Router. Is that correct? Here, Admin Router needs to be fully aware of the (scheme, hostname, port) tuple that's used by a user agent to reach the first proxy in the chain.

The patch you propose, I believe, implicitly relies on two rather adventurous assumptions:

* `$scheme` matches the scheme used to address the previous proxy: I don't think this is a safe assumption, it probably only works for you by coincidence.
* `$http_host` contains host _and_ port information: w/o clearly defining requirements on how exactly the previous proxy has to populate the HOST header, this is also not a safe assumption.

In order to tackle this properly, we need to have more clarity on the requirements and then precisely define the behavior that what we want to have. This of course translates into requirements we have for every previous proxy (we may e.g. require that the previous proxy _must_ retain the original request URL).

I hope this illustrates that we need to better define the requirements:

* the previous proxy may or may not set the HOST header to contain its host name _and_ its port.
* the previous proxy may or may not set the HOST header to contain its host name and an additional header to contain its port.
* the previous proxy may or may not retain the original request URL (containing the original scheme, host name, and port).

In Admin Router we have many different options to retain the original scheme, host, port. For instance, we may want to use `$host` in the `proxy_redirect` replacement string (quote from [docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_host)):

> in this order of precedence: host name from the request line, or host name from the “Host” request header field, or the server name matching a request

On the other hand, a simple and flexible setup could be
* require all previous proxies to retain the original request line
* "manually" extract scheme, host, port from that request line

So, we should do a better job in defining requirements and then implement a corresponding solution.

alberts:
We're going to publish a doc with how to properly configure haproxy for this instead.

Internal ticket: https://mesosphere.atlassian.net/browse/DCOS-9559

PR #18, opened 2016-10-06
validate jwt for 3dt and pkgpanda API

darkonie:
call `auth.validate_jwt_or_exit()` on all masters for 3dt and pkgpanda API

### DIFF
--- a/common/server.conf
+++ b/common/server.conf
@@ -8,19 +8,3 @@ ssl_ciphers EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:
 ssl_prefer_server_ciphers on;
 
 server_name dcos.*;
-
-location /system/health/v1 {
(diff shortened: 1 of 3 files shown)


spahl:
:ship: 

@alberts can you also have a look? Tests in DC/OS pass

branden:
To prevent duplicating that config between the masters and agents, I think this could be done by keeping the pkgpanda and 3DT locations (as well as their `access_by_lua 'auth.validate_jwt_or_exit()';`) in `common/server.conf` and adding a location to `nginx.agent.conf` for `/acs/api/v1/auth/` that does a `proxy_pass` to `master.mesos`. This is why I put `auth.lua` under `common` when I assembled the agent adminrouter -- both masters and agents ought to be able to use it.

branden:
I take that back, it looks like we'd need to put `/var/lib/dcos/dcos-oauth/auth-token-secret` on agents in order to handle the JWT. I should probably move `auth.lua` to `master/` :\

PR #19, opened 2016-11-01
DCOS-9269 Standarizing open and ee AR

vespian:
Testing stuff here.


nginx.master.conf
```diff
@@ -14,24 +18,24 @@ http {
         server master.mesos:8080;
     }
 
+    upstream cosmos {
+        server localhost:7070;
```
alberts: let's use 127.0.0.1 here too.

vespian: :+1:

PR #20, opened 2016-11-08
Add metrics upstream

malnick:


malnick:
@spahl This is ready for review.

vespian:
Look good!

Would it be a problem to also add `proxy_set_header X-Forwarded-Proto $scheme;` to the location stanzas just as a good measure ? It will most probably not be used by the upstream, but it would be nice if we started thinking about a common set of headers we always pass to upstreams.

Thanks!

malnick:
Can do!

malnick:
@vespian This is updated.

nginx.agent.conf
```diff
@@ -31,5 +35,12 @@ http {
             proxy_pass http://pkgpanda/;
             proxy_redirect http://$http_host/ /pkgpanda/;
         }
+
+        location /system/metrics {
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+						proxy_set_header X-Forwarded-Proto $scheme;
```
vespian: Could you please change the tabs into spaces here and down below ?

malnick:
@vespian This should be ready now.

PR #21, opened 2016-11-15
add dcos-log reverse proxy

darkonie:
`dcos-log` https://github.com/dcos/dcos-log is an HTTP server over
systemd journal. dcos-log supports server-sent events for live log pushing
that is why we should disable nginx buffering. dcos-log will be deployed as a
pkgpanda package on all nodes: masters and agents/public agents.

In open DC/OS we are using boolean authorization. We should allow access to
service to any user who is authenticated.

### DIFF
--- a/common/http.conf
+++ b/common/http.conf
@@ -34,3 +34,7 @@ init_by_lua '
 upstream pkgpanda {
     server unix:/run/dcos/pkgpanda-api.sock;
 }
+
+upstream log {
+    server unix:/run/dcos/dcos-log.sock;
+}
--- a/master/agent.lua
+++ b/master/agent.lua
@@ -9,7 +9,10 @@ end
 for _, agent in ipairs(state["slaves"]) do
     if agent["id"] == ngx.var.agentid then
         local split_pid = agent["pid"]:split("@")
-        ngx.var.agentaddr = split_pid[2]
+        local host_port = split_pid[2]:split(":")
+        ngx.var.agentaddr = host_port[1]
+        ngx.var.agentport = host_port[2]
+
         ngx.log(
             ngx.DEBUG, "agentid / agentaddr:" ..
             ngx.var.agentid .. " / " .. ngx.var.agentaddr
--- a/nginx.agent.conf
+++ b/nginx.agent.conf
@@ -31,5 +31,21 @@ http {
             proxy_pass http://pkgpanda/;
             proxy_redirect http://$http_host/ /pkgpanda/;
         }
+
+        location /system/logs/v1/ {
+            proxy_buffering off;
+            proxy_cache off;
+
+            # make event source work with nginx
+            # http://stackoverflow.com/questions/13672743/eventsource-server-sent-events-through-nginx
+            proxy_set_header Connection '';
+            proxy_http_version 1.1;
+            chunked_transfer_encoding off;
+
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            proxy_set_header Host $host;
+            proxy_set_header X-Forwarded-Proto $scheme;
+            proxy_pass http://log/;
+        }
     }
 }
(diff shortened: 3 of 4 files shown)


common/server.conf
```diff
@@ -8,3 +8,17 @@ ssl_ciphers EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:
...
+    access_by_lua 'auth.validate_jwt_or_exit()';
+    proxy_buffering off;
+    proxy_cache off;
+
+    proxy_set_header Connection '';
+    proxy_http_version 1.1;
+    chunked_transfer_encoding off;
+
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header Host $host;
```
vespian: Could you please also add here `proxy_set_header X-Forwarded-Proto $scheme` ? It will most probably not be used by the upstream, but it would be nice if we started thinking about a common set of headers we always pass to upstreams.

common/server.conf
```diff
@@ -8,3 +8,17 @@ ssl_ciphers EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:
 ssl_prefer_server_ciphers on;
 
 server_name dcos.*;
+
+location /system/logs/v1/ {
+    access_by_lua 'auth.validate_jwt_or_exit()';
+    proxy_buffering off;
+    proxy_cache off;
+
+    proxy_set_header Connection '';
```
vespian: Could you please add some comment here regarding why we want to keep upstream connection up after request ? We can just copy the text from the PR description and i.e. link this SO for more details: http://stackoverflow.com/questions/13672743/eventsource-server-sent-events-through-nginx.

master/logging.lua
```diff
@@ -0,0 +1,17 @@
+local util = require "master.util"
+
+local state = util.mesos_get_state()
+if state == nil then
+    ngx.status = ngx.HTTP_SERVICE_UNAVAILABLE
+    ngx.say("503 Service Unavailable: invalid Mesos state.")
+    return ngx.exit(ngx.HTTP_SERVICE_UNAVAILABLE)
+end
+for _, agent in ipairs(state["slaves"]) do
+    if agent["id"] == ngx.var.agentid then
```
vespian: I diffed logging.lua with agent.lua and:
```
 for _, agent in ipairs(state["slaves"]) do
     if agent["id"] == ngx.var.agentid then
-        local split_pid = agent["pid"]:split("@")
-        ngx.var.agentaddr = split_pid[2]
-        ngx.log(
-            ngx.DEBUG, "agentid / agentaddr:" ..
-            ngx.var.agentid .. " / " .. ngx.var.agentaddr
-            )
+        ngx.var.agentaddr = agent["hostname"]
         return
     end
 end
```

Could you please give me some context on this ?

darkonie: we should probably make it generic so auth.lua and logging.lua can reuse the same bits. I am just not quite sure what would be the best way of doing it.

The idea here is simple, user passed a nodeID and nginx looks up in a map `mesos id = node ip` and does the reverse proxy to agent adminrouter.

vespian: > we should probably make it generic so auth.lua and logging.lua can reuse the same bits
Do you mean `agent.lua` instead of `auth.lua` ?

> I am just not quite sure what would be the best way of doing it.
We can make this lua code set two variables: `$agentip` and `$agentport`. The `/(slave|agent)/` endpoint can do `proxy_pass http://$agentip:$agentport` while you will just `proxy_pass http://$agentip:61001/system/logs/v1/$url$is_args$query_string`. We would have to double check with core people if we can rely on `agent["pid"]` always following `<type>@<ip>:<port>` format.

common/server.conf
darkonie: added

common/server.conf
darkonie: added

darkonie:
i had to duplicate `location /system/logs/v1/` in `nginx.agent.conf` and `nginx.master.conf` because agent seems to not treat `access_by_lua 'auth.validate_jwt_or_exit()';` correctly, besides it is consistent with the rest of endpoints eg `/system/health/v1` and pkgpanda.

the change to `agent.lua` is working! both `proxy_pass http://$agentip:$agentport` and `proxy_pass http://$agentip:61001` work.

PR #24, opened 2016-11-17
change dcos-log endpoints

darkonie:
/system/logs/v1 -> /system/v1/logs/v1
/system/logs/v1/agent/<agent_id>/endpoint -> /system/v1/logs/v1/agent/<agent_id>/endpoint

### DIFF
--- a/nginx.agent.conf
+++ b/nginx.agent.conf
@@ -36,26 +36,15 @@ http {
             proxy_redirect http://$http_host/ /pkgpanda/;
         }
 
-        location /system/logs/v1/ {
-            proxy_buffering off;
-            proxy_cache off;
+        location /system/v1/logs/v1/ {
+            include common/server-sent-events.conf;
+            include common/proxy-headers.conf;
 
-            # make event source work with nginx
-            # http://stackoverflow.com/questions/13672743/eventsource-server-sent-events-through-nginx
-            proxy_set_header Connection '';
-            proxy_http_version 1.1;
-            chunked_transfer_encoding off;
-
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-            proxy_set_header Host $host;
-            proxy_set_header X-Forwarded-Proto $scheme;
             proxy_pass http://log/;
         }
-        
+
         location /system/metrics {
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-            proxy_set_header X-Forwarded-Proto $scheme;
-            proxy_set_header Host $host;
+            include common/proxy-headers.conf;
             proxy_pass http://metrics;
         }
     }
--- a/nginx.master.conf
+++ b/nginx.master.conf
@@ -266,52 +266,41 @@ http {
             proxy_redirect http://$http_host/ /pkgpanda/;
         }
 
-        location /system/logs/v1/ {
+        location /system/v1/logs/v1/ {
             access_by_lua 'auth.validate_jwt_or_exit()';
-            proxy_buffering off;
-            proxy_cache off;
-
-            # make event source work with nginx
-            # http://stackoverflow.com/questions/13672743/eventsource-server-sent-events-through-nginx
-            proxy_set_header Connection '';
-            proxy_http_version 1.1;
-            chunked_transfer_encoding off;
-
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-            proxy_set_header Host $host;
-            proxy_set_header X-Forwarded-Proto $scheme;
+            include common/server-sent-events.conf;
+            include common/proxy-headers.conf;
             proxy_pass http://log/;
         }
 
-        location ~ ^/system/logs/v1/(slave|agent)/(?<agentid>[0-9a-zA-Z-]+)(?<url>.*)$ {
+        location ~ ^/system/v1/agent/(?<agentid>[0-9a-zA-Z-]+)/logs/v1/(?<url>.*)$ {
             access_by_lua 'auth.validate_jwt_or_exit()';
             set $agentaddr '';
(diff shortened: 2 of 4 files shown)


nginx.master.conf
```diff
@@ -283,7 +283,7 @@ http {
             proxy_pass http://log/;
         }
 
-        location ~ ^/system/logs/v1/(slave|agent)/(?<agentid>[0-9a-zA-Z-]+)(?<url>.*)$ {
+        location ~ ^/system/v1/logs/v1/(slave|agent)/(?<agentid>[0-9a-zA-Z-]+)(?<url>.*)$ {
```
darkonie: does this look reasonable?

darkonie: yes, we had this discussion here. We will change `/system/health` to `/system/v1/health/v1`

darkonie: but only after 1.9 release

nginx.master.conf
```diff
@@ -266,47 +266,48 @@ http {
...
-            proxy_buffering off;
-            proxy_cache off;
+            include common/server-sent-events.conf;
 
-            # make nginx with with server sent events
-            # http://stackoverflow.com/questions/13672743/eventsource-server-sent-events-through-nginx
-            proxy_set_header Connection '';
-            proxy_http_version 1.1;
-            chunked_transfer_encoding off;
+            proxy_set_header Host $http_host;
```
cmaloney: would be nice to pull the proxy_set_header into an include file if we can rather than duplicating it.

darkonie: added `common/proxy-headers.conf`

nginx.agent.conf
```diff
@@ -36,15 +36,8 @@ http {
...
-
-            # make event source work with nginx
-            # http://stackoverflow.com/questions/13672743/eventsource-server-sent-events-through-nginx
-            proxy_set_header Connection '';
-            proxy_http_version 1.1;
-            chunked_transfer_encoding off;
+        location /system/v1/logs/v1/ {
+            include common/server-sent-events.conf;
 
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```
cmaloney: `include common/proxy-headers.conf`?

cmaloney: Same goes for the metrics endpoint. Seems like the X-Real-IP and the like are good things to include.

nginx.master.conf
```diff
@@ -266,47 +266,41 @@ http {
...
-            proxy_cache off;
-
-            # make event source work with nginx
-            # http://stackoverflow.com/questions/13672743/eventsource-server-sent-events-through-nginx
-            proxy_set_header Connection '';
-            proxy_http_version 1.1;
-            chunked_transfer_encoding off;
+            include common/server-sent-events.conf;
 
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```
cmaloney: `include common/proxy-headers.conf`?

nginx.agent.conf
darkonie: fixed

nginx.master.conf
darkonie: fixed

vespian:
Just my post-merge two cents, feel free to ignore ;)

Problems I see here can be nicely summarized by this xkcd comic strip: https://xkcd.com/927/ 

I think that introducing proxy-headers.conf adds just one more way we configure headers for proxies, quick search to name a few:
* https://github.com/mesosphere/adminrouter/blob/c2d52cf3f58751d05493ed4a5c9af43db1fe9e18/common/server.conf#L10
* https://github.com/mesosphere/adminrouter/blob/37457dcb79092ef0bea39c7f78fdc34cc31b12bb/nginx.agent.conf#L35
* https://github.com/dcos/adminrouter/blob/5e23b06be6c061b394d2123424675c542072a6f2/nginx.agent.conf#L31

It would have made more sense if we refactored *all* endpoints to include this file, or none and just keept it simple. On top of that - many people are working on this repository, the more indirection layers we add (aka includes), the harder for them will be to understand the configuration and make right changes. I am hoping to come up with something better than copypasting while working on open-ee AR merge but this is still WIP.

When it comes to the extra api versioning, very similar arguments may be brought up, but here we also have a situation when a team is responsible for some endpoint ('/system/'). I do not know how we want to see AR api in the future - whether we want to give feature teams some degree of freedom and let them add endpoints that suit them best or is AR api consistency our main goal. Both are fine for me :)

cmaloney:
From my perspective, definitely _should_ use the same proxy include everywhere. The goal with moving there is that people keep leaving out some headers sometimes, and I'd _much_ rather have consistent behavior with all the proxy header fields, rather than a hodge-podge where different backend services get different things.

As far as versioning:
From the customer's point of view, DC/OS should have one API. That different components have different teams working on them ideally is not something the API user should need to be aware of. APIs shouldn't expose their internal workings, just the API that clients are looking to speak.

Ideally we'd have a strong standard for what all the APIs return. As things likely move more GraphQL based that will probably happen some, although for now we get whatever each team implemented with their own bespoke rest implementations, all of which vary slightly... The amount of code in the DC/OS CLI to deal with the different behaviors of our different endpoints/components is a bit crazy.

vespian:
> The goal with moving there is that people keep leaving out some headers sometimes, and I'd much rather have consistent behavior with all the proxy header fields, rather than a hodge-podge where different backend services get different things.

Totally :) We just disagree how we can reach the sweat spot. Includes add another layer of indirection, the more layers the more it is going to be difficult for people to wrap their heads around it. I was thinking about having tests bundled with AR that will enforce correct headers. This is WIP. For now I just wanted to point out that it's better to be consistent and either change all the endpoints or none.

> ...for now we get whatever each team implemented with their own bespoke rest implementations, all of which vary slightly...

:+1:

PR #25, opened 2016-11-18
Use /sytsem/v1/metrics base route for metrics serivce

malnick:


alberts:
please squash

nginx.master.conf
```diff
@@ -297,11 +297,13 @@ http {
 
             proxy_pass http://$agentaddr:61001/system/v1/$url$is_args$query_string;
         }
-
-        location /system/metrics {
+        
+        location /system/v1/metrics/ {
             access_by_lua 'auth.validate_jwt_or_exit()';
-            include common/proxy-headers.conf;
```
alberts: why did we stop using this include? is anything else still using this file?

alberts:
also, please fix typo in commit message

nginx.agent.conf
```diff
@@ -42,10 +42,12 @@ http {
 
             proxy_pass http://log/;
         }
-
-        location /system/metrics {
-            include common/proxy-headers.conf;
-            proxy_pass http://metrics;
+        
+        location /system/v1/metrics/ {
```
alberts: we're renaming endpoints... are we sure no customer might be using there?

malnick: These are new services, so no customers are changing them. We want to version the /system/ URI to make it feel more like a single API.

malnick:
@alberts Squashed, spelling corrected.

nginx.agent.conf
spahl: Correct

PR #26, opened 2016-11-22
Switch all proxy sites to use the same set of common headers

cmaloney:
Unifies the proxying behavior for everywhere we were doing proxy_set_headers. Takes us from 45 proxy_set_header instances to 7. Overall goal is that we have less unnecessary variation inside the product.

Might break things, might not. Let's find out!

cc: @vespian

### DIFF
--- a/nginx.agent.conf
+++ b/nginx.agent.conf
@@ -21,16 +21,12 @@ http {
         listen 61001 default_server;
 
         location /system/health/v1 {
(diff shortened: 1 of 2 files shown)


cmaloney:
testing DC/OS PR: https://github.com/dcos/dcos/pull/982

cmaloney:
DC/OS tests all passed, so this probably didn't break anything / is safe to merge.

PR #28, opened 2016-11-29
Update socket paths to reflect per-role socket allocation

malnick:


malnick:
@alberts @spahl This is ready for review.

PR #29, opened 2016-11-30
Disable buffering to mesos agent API and use HTTP/1.1

malnick:
This is required for debugging to work.


nginx.master.conf
```diff
@@ -153,6 +153,10 @@ http {
             proxy_set_header X-Forwarded-Proto $scheme;
 
             proxy_pass http://$agentaddr:$agentport;
+
+						proxy_http_version 1.1;
```
klueska: Looks like the indentation is off here.

PR #30, opened 2016-12-07
Don't override host in webui_url.

alberts:
This was breaking the Elastic framework, which was running its UI in a separate Kibana container, that could be scheduled on a different host than the framework itself.


jgehrcke:
Let's test ArangoDB before merging.

alberts:
@vespian sadly, the original commits have no clues about why we did this. We'll test as many frameworks as possible before proceeding.

PR #37, opened 2017-01-17
Cache refactoring & adding "marathon leader" related stuff

vespian:
Issues affected:
* DCOS-10223
* DCOS-13189

This PR:
* removes Dockerfile code that moves /tmp to /dev/shm directory. It was causing more trouble than it was worth it.
* small cleanup and spelling/formatting/logging errors
* refactors LineBufferFilter in a way that permits complex search criteria (regexes, occurrence counts)
* change default log level for AR from Notice to Info, as otherwise, it's not possible to reliably test some parts of the code (-> cache). Using `-g loglevel stderr INFO` command line option does not solve this problem as it only adds another log source, thus making AR print all >=NOTICE level messages twice
* Fix bug in Mocker - it should wait for all the endpoints to start before continuing
* make marathon and Mesos mocker endpoints more capable, including but not limited to:
  * marathon leader endpoint
  * simulated task in marathon
  * simulated extra agent in Mesos
* add simulating brokerage and timing out for endpoints
* fix a bug for `/service/<service-id>/` endpoint - if service was not found in cache/marathon 500 error was returned instead of 404
* backport DEFAULT_SCHEME environment variable from EE. This helps to simplify cache code in regards to EE vs. Open repo split
* refactor AR cache:
  * better handle errors, including but not limited to:
    * only one of upstreams is down
    * introduce `cache_max_age_soft_limit` and `max_age_hard_limit` concepts
  * support for marathon-leader caching
  * rename mesosstatecache to just `cache` as it performs multiple roles now
  * make first poll period, poll period, and cache expiration configurable via env variables in order to make the code testable
  * unify the interface by which cache is accessed: `get_cache_entry(name)`
  * add tests that cover all the changes
* add Logging api /leader/* endpoints and tests

Flake8 passes, tests pass - ATM we have ~100 of them.


vespian:
@jgehrcke @alberts @darkonie @malnick  Please have a look.Thanks!

master/cache.lua
```diff
@@ -57,19 +85,27 @@ local function request(host, port, path)
...
     ngx.log(ngx.NOTICE, "Cache Marathon app state")
-    local appsRes, err = request("127.0.0.1", 8080, "/v2/apps?embed=apps.tasks&label=DCOS_SERVICE_NAME")
+    local appsRes, err = request("127.0.0.1", 8080,
+								 "/v2/apps?embed=apps.tasks&label=DCOS_SERVICE_NAME",
+                                 false)
 
     if err then
         ngx.log(ngx.NOTICE, "Marathon app request failed: " .. err)
+        if not cache_data("svcapps", nil) then
+            ngx.log(ngx.ERR, "Invalidating Marathon apps cache failed")
```
jgehrcke: What follows is only is an important point if we do this kind of cache invalidation at all: it seems that setting the cache value to `nil` without adjusting the corresponding "last refresh time" value seems inconsistent (but I might have missed something in the code).

test-harness/tests/test_cache.py
```diff
@@ -0,0 +1,476 @@
+# -*- coding: utf-8 -*-
```
vespian: 👍

README.md
```diff
@@ -109,9 +108,9 @@ Mocker takes care of simulating DC/OS HTTP endpoints that Admin Router uses. It'
 basically just a thin management layer on top of multiple python-based
 HTTP servers, each one of them mocking out particular component/upstream of DC/OS.
 
-It exposes the `.send_command()` method that allows to reconfigure endpoints
-the way the tests need it. It basically calls a getattr on the endpoint
-instance in order to execute given function and pass it attributes as specified
+It exposes the `.send_command()` method that allows to endpoint reconfiguration
```
vespian: 👍

common/main.conf
```diff
@@ -1,13 +1,18 @@
...
 
 
 # Make env vars accessible from within Lua modules.
 env SECRET_KEY_FILE_PATH;
 env AUTH_ERROR_PAGE_DIR_PATH;
 env OAUTH_CLIENT_ID;
 env OAUTH_AUTH_REDIRECTOR;
+env DEFAULT_SCHEME;
+
+env FIRST_POLL_PERIOD_SECONDS;
```
vespian: true, fixed.

common/main.conf
```diff
@@ -1,13 +1,18 @@
...
+env POLL_PERIOD_SECONDS;
+env CACHE_EXPIRATION_SECONDS;
```
vespian: +1 to CACHE_ - seems nice, but I am not sure if adding _SECONDS suffix is a good idea/it's rarely used in practice. It seems redundant - units should be documented in the code.

I have renamed it the way you suggested plus added a comment. I do not think it's worth having longer discussion about it :)

docker/Dockerfile
```diff
@@ -109,11 +108,16 @@ ENV AUTH_ERROR_PAGE_DIR_PATH=${AR_BIN_DIR}/nginx/conf/errorpages
 COPY ./ca.crt /run/dcos/pki/CA/certs/ca.crt
 COPY ./adminrouter.crt /run/dcos/pki/tls/certs/adminrouter.crt
 COPY ./adminrouter.key /run/dcos/pki/tls/private/adminrouter.key
-COPY ./adminrouter-listen-master.conf /opt/mesosphere/etc/adminrouter-listen-master.conf
-COPY ./adminrouter-listen-agent.conf /opt/mesosphere/etc/adminrouter-listen-agent.conf
 COPY ./adminrouter-redirect-http-https.conf /opt/mesosphere/etc/adminrouter-redirect-http-https.conf
 COPY ./adminrouter-upstreams.conf /opt/mesosphere/etc/adminrouter-upstreams.conf
 
+# These file's contents differ from DC/OS counterpart, as the IP addresses
```
vespian: `The contents of this file differ from the one that is shipped with DC/OS:`

master/cache.lua
```diff
@@ -5,17 +5,45 @@ local http = require "resty.http.simple"
 
 local _M = {}
 
+-- In order to make caching code testable, these constants need to be
+-- configurable/exposed through env vars.
+local FIRST_POLL_PERIOD_SECONDS = os.getenv("FIRST_POLL_PERIOD_SECONDS")
+if FIRST_POLL_PERIOD_SECONDS == nil then
+    FIRST_POLL_PERIOD_SECONDS = 2
+    ngx.log(ngx.DEBUG, "FIRST_POLL_PERIOD_SECONDS not set")
```
vespian: 👍

master/cache.lua
```diff
@@ -43,7 +71,7 @@ local function request(host, port, path)
         return nil, err
     end
 
-    if res.status ~= 200 then
+    if res.status ~= 200 and accept_404_reply and res.status ~= 404 then
```
vespian: Actually there are two bugs here - one that you found and the other in the LineBufferFilter class that prevented test from detecting this one :(

Thanks a lot for catching it!

master/cache.lua
```diff
@@ -57,19 +85,27 @@ local function request(host, port, path)
...
 
 
-local function fetch_and_cache_state_marathon()
+local function fetch_and_store_marathon_apps()
     -- Access Marathon through localhost.
     ngx.log(ngx.NOTICE, "Cache Marathon app state")
-    local appsRes, err = request("127.0.0.1", 8080, "/v2/apps?embed=apps.tasks&label=DCOS_SERVICE_NAME")
+    local appsRes, err = request("127.0.0.1", 8080,
+								 "/v2/apps?embed=apps.tasks&label=DCOS_SERVICE_NAME",
+                                 false)
```
vespian: 👍

master/cache.lua
```diff
@@ -151,50 +187,133 @@ local function fetch_and_cache_state_marathon()
...
-    local success = cache_data("svcapps", svcApps)
-    if success then
-        cache_data("last_fetch_time_marathon", time_fetched)
+    return
+end
+
+
+local function fetch_and_store_marathon_leader()
+    -- Fetch state JSON summary from Mesos. If successful, store to SHM cache.
+    -- Expected to run within lock context.
```
vespian: +1

master/cache.lua
```diff
@@ -151,50 +187,133 @@ local function fetch_and_cache_state_marathon()
...
+    local time_now = ngx.now()
+    if cache_data("mesosstate_last_refresh", time_now) then
+        ngx.log(ngx.INFO, "Mesos state cache has been successfully updated")
+    end
+
+    return
+end
+
+
+local function refresh_needed(ts_name)
```
vespian: time-stamp name. I am not attached to this name, if you have something better in mind then go ahead.

master/cache.lua
```diff
@@ -286,77 +399,49 @@ function _M.periodically_poll_mesos_state()
...
-            )
-            return nil
-        end
-        ngx.log(
-            ngx.NOTICE,
-            "Mesos state not available in cache yet. Fetch it."
-        )
-        refresh_mesos_state_cache()
-        return get_state_summary(true)
+    local entry_json = cache:get(name)
```
vespian: yes

master/cache.lua
```diff
@@ -151,50 +187,133 @@ local function fetch_and_cache_state_marathon()
...
+    if cache_data("mesosstate_last_refresh", time_now) then
+        ngx.log(ngx.INFO, "Mesos state cache has been successfully updated")
+    end
+
+    return
+end
+
+
+local function refresh_needed(ts_name)
+    -- Handle special case of first invocation.
```
vespian: +1

master/mleader.lua
```diff
@@ -0,0 +1,21 @@
+local cache = require "master.cache"
```
vespian: `marathon-leader.lua`

test-harness/modules/generic_test_code.py
```diff
@@ -9,6 +9,17 @@
 
 
 def generic_unknown_user_is_forbidden_test(ar, auth_header, path):
+    """Test if invalid user is forbidden access
```
vespian:
```
def generic_unauthed_user_is_forbidden_test(ar, auth_header, path):
    """Test if unauthorized/unauthenticated user is forbidden access
```

vespian: user stems from uid claim in JWT token - it's a random string unknown to the system.

test-harness/tests/test_master.py
```diff
@@ -134,3 +137,78 @@ def test_if_upstream_headers_are_correct(self,
...
+        ('',
+         ''),
+    ])
+    def test_if_upstream_request_is_correct(self,
+                                            master_ar_process,
+                                            valid_user_header,
+                                            sent,
+                                            expected,
+                                            endpoint_type):
+
```
vespian: It would be wonderfull to have this as a flake8 rule/check - we could agree on something and have it enforced across all repos.

Fixed.

vespian: Niceness restored.

test-harness/tests/open/test_master.py
```diff
@@ -158,15 +158,15 @@ def test_if_valid_auth_attempt_is_logged_correctly(
...
+        assert lbf.all_found
 
     def test_if_invalid_auth_attempt_is_logged_correctly(
             self, master_ar_process, valid_jwt_generator):
         # Create some random, unique user that we can grep for:
         uid = 'some_random_string_abc1251231143'
 
-        filter_string = 'validate_jwt_or_exit(): User not found: `{}`'.format(uid)
-        lbf = LineBufferFilter(filter_string,
+        filter_regexp = 'validate_jwt_or_exit\(\): User not found: `{}`'.format(uid)
```
vespian: Discussed in person. When the future comes, we will be ready.

test-harness/tests/test_cache.py
```diff
@@ -0,0 +1,476 @@
...
+        with GuardedAR(ar):
+            resp = requests.get(url,
+                                allow_redirects=False,
+                                headers=valid_user_header)
+            assert resp.status_code == 404
+
+            mocker.send_command(endpoint_id='http://127.0.0.1:8080',
+                                func_name='enable_nginx_task')
+
+            time.sleep(poll_period)
```
vespian: It does not matter, but not for the reason you mentioned - the refresh *has* to happen before we execute mocker.send_command() as we want to observe the the registering new marathon apps data.

Timeline is as follows:
* nginx starts
* cache updates are scheduled:
  * CACHE_FIRST_POLL_DELAY_SECONDS by default is 2
  * CACHE_POLL_PERIOD_SECONDS is overriden to 4
  * CACHE_EXPIRATION_SECONDS is overriden to 3
* first request is made. It forces cache refresh that sets `marathonleader_last_refresh` if it happens before first poll, it uses the cache data if it happens after the first poll. This way or another we are guaranteed to have some initial data in the cache before we execute the command and we are 100% certain what this data is.
* we change the mesos state data.
* poll_period > cache_expiry, so we are guaranteed to have the cache expired by the time we perform next query. It will be refreshed by either a timer refresh or request-enforced refresh.
* we inspect response data to see if the mesos state change was reflected in the cache.

Hope it helps.

master/cache.lua
```diff
@@ -151,50 +187,133 @@ local function fetch_and_cache_state_marathon()
        ::continue::
     end
 
-    svcApps = cjson_safe.encode(svcApps)
-    ngx.log(ngx.DEBUG, "storing services " .. svcApps)
+    svcApps_json = cjson_safe.encode(svcApps)
+
+    ngx.log(ngx.DEBUG, "Storing Marathon services data to SHM.")
+    if not cache_data("svcapps", svcApps_json) then
+        ngx.log(ngx.ERR, "Storing marathon apps cache failed")
```
vespian: `apps` it is - there was only one occurence of `services`.

common/main.conf
```diff
@@ -1,13 +1,18 @@
 # Log notice level and higher (e.g. state cache
 # emits useful log messages on notice level).
-error_log stderr notice;
+error_log stderr info;
```
vespian: Discussed in person. The conclusion (correct me if I am wrong) is that it does not change much the amount of data logged and that the issue will be addressed as soon as it's technically possible (we will have to drop MesosDNS and switch to direct API queries which is not trivial and has some dependencies on its own).

common/server.conf
```diff
@@ -1,6 +1,3 @@
-ssl_certificate common/snakeoil.crt;
```
alberts: are we sure about this?

vespian: We are not removing it, it just got moved into a separate file and DC/OS package `build` script will have to be adjusted accodingly. Check the Dockerfile changes.

vespian: The reason for this change is two-fold:
* AR by default listens on 0.0.0.0:80, this makes it impossible to mock out  i.e. marathon leader proxying
* all listen and ssl_* related stuff in EE is in a separate file as well. Unifying the approach will make the merge a bit easier.

vespian:
@jgehrcke Thank you very much for the time you have spent giving this detailed feedback! I appreciate it :)

This is ready for the second pass! Changes include:
* make all cache timing variables exposed as ENV vars for testing
* implement `max_age_(soft|hard)_limit` vars and the behaviour we discussed in general
* fixed cache invalidation bug - no more nuking cache
* all the other small fixes you pointed out except for the https://github.com/dcos/adminrouter/pull/37#discussion_r97664184 comment

Cheers!

vespian:
@jgehrcke @gpaul I have separated the changes we discussed into a separate PR based against this one: https://github.com/dcos/adminrouter/pull/41

test-harness/modules/generic_test_code.py
```diff
@@ -8,7 +7,52 @@
 log = logging.getLogger(__name__)
 
 
-def generic_unknown_user_is_forbidden_test(ar, auth_header, path):
+def ping_mesos_agent(ar,
```
mhrabovcin: How/where is this function used?

vespian: Very often when we play with networking, we issue `ping 8.8.8.8` to test network connectivity. `ping_mesos_agent` has the same role for testing cache - it indirectly tests if Mesos statecache was cached properly and given agent is reachable. It abstracts the act of:
* forming the right URL
* issuing the request
* checking if the response was sent by correct upstream

It will most probably have many more uses as the number of tests keeps growing, though.

I hope it helps :)

mhrabovcin: Got it! I just don't see this function being actually used anywhere in the code.

vespian: This is because GH folds files which produce too much diff output. Try expanding `test-harness/tests/test_cache.py ` and you will find a lot of uses of this function.

master/service.lua
```diff
@@ -41,3 +52,7 @@ for _, framework in ipairs(state["frameworks"]) do
     end
   end
 end
+
+ngx.status = ngx.HTTP_NOT_FOUND
+ngx.say("404 Not Found: service " .. ngx.var.serviceid .. " unknown.")
```
mhrabovcin: Can we add timestamp when was the cache refreshed? Related to https://mesosphere.atlassian.net/browse/DCOS-13596

vespian: will do as a separate PR, on top of this one. Will send a link soonish.

gpaul:
All of this looks good - except this
- introduce cache_max_age_soft_limit and max_age_hard_limit concepts

This is a change of behaviour in a PR that is intended to refactor which in general means one is entering tricky territory.

In this case, I don't understand why we have three levels of cache? It causes all the weirdness in https://github.com/dcos/adminrouter/pull/41 and I don't see what it is doing here?

vespian:
@jgehrcke Thanks for another pass.

> is there a particular reasoning behind the log level choice between INFO and NOTICE? If there are no strong reasons for using both levels in this code, I think we're better off if we stick to just one of them.

For example:
http://ux.stackexchange.com/a/82225
http://stackoverflow.com/a/36656073

I like this explanation a lot:
```
notice:
  Normal but significant [notable] condition.
  "httpd: caught SIGBUS, attempting to dump core in ..."

info:
  Informational [and unnotable].
  ["Server has been running for x hours."]
```

> Now that we handle the 'heavily outdated' case, it's time to reason about the specific value. Clearly, we want this to be useful. I believe a 3 day period might be a bit too long before throwing errors (an extreme point of view would be: does this ever take effect in the real world? If it does not, why did we build it? So, we want this to take effect, and the value must be set accordingly). Would appreciate if you could talk through that value choice with @alberts.

Let's do it as a separate PR where we can discuss it in detail.

The current value is intentionally "too long" so that we get the behaviour similar to the original one and at the same time we introduce some very basic sanity-check: "Hey - your cluster is missing Mesos updated for 3 days already!".

master/marathon-leader.lua
```diff
@@ -0,0 +1,21 @@
+local cache = require "master.cache"
+
+local mleader = cache.get_cache_entry("marathonleader")
+if mleader == nil then
+    ngx.status = ngx.HTTP_SERVICE_UNAVAILABLE
+    ngx.say("503 Service Unavailable: unable to obtain Marathon leader cache")
```
vespian: Of course, we can unify it - I did not want to change too many things in this already big PR.

I standardised on `503 Service Unavailable: cache state is invalid`.

master/cache.lua
```diff
@@ -0,0 +1,465 @@
...
+-- configurable/exposed through env vars.
+--
+-- Values assigned to these variable need to fufil following condidtion:
+--
+-- CACHE_FIRST_POLL_DELAY << CACHE_EXPIRATION < CACHE_POLL_PERIOD < CACHE_MAX_AGE_SOFT_LIMIT < CACHE_MAX_AGE_HARD_LIMIT
+--
+-- CACHE_BACKEND_REQUEST_TIMEOUT << CACHE_REFRESH_LOCK_TIMEOUT
+--
+-- All are in units of seconds. Below are the defaults:
+local _TIMERS = {}
```
vespian: True, _TIMERS does not really fit here.

The problem I am facing here is that it's not CONFIG config either, it is a set of default values for various timers. Naming it DEFAULT_VALUES_FOR_TIMERS seems weird.

Would _DEFAULTS work for you? The names of variables here are pretty descriptive already.

master/cache.lua
```diff
@@ -0,0 +1,465 @@
...
+            path = path,
+            timeout = _TIMERS.CACHE_BACKEND_REQUEST_TIMEOUT * 1000,
+        }
+    )
+
+    if not res then
+        return nil, err
+    end
+
+    if res.status ~= 200 and (accept_404_reply and res.status ~= 404 or not accept_404_reply) then
```
vespian: I added one more nesting level. `accept_404_reply and res.status ~= 404 or not accept_404_reply` will have to stay as-is unless we want to duplicate code.

master/cache.lua
```diff
@@ -0,0 +1,465 @@
...
+end
+
+
+local function fetch_and_store_marathon_leader()
+    -- Fetch Marathon leader address. If successful, store to SHM cache.
+    -- Expected to run within lock context.
+    local mleaderRes, err = request("127.0.0.1", 8080, "/v2/leader", true)
+
+    if err then
+        ngx.log(ngx.NOTICE, "Marathon leader request failed: " .. err)
```
vespian: Nope, changed to WARN.

master/cache.lua
```diff
@@ -0,0 +1,465 @@
...
+    -- We need to translate 404 reply into a JSON that is easy to process for
+    -- endpoints:
+    -- https://mesosphere.github.io/marathon/docs/rest-api.html#get-v2-leader
+    local res_body
+    if mleaderRes.status == 404 then
+        -- Just a hack in order to avoid using gotos - create a substitute JSON
+        -- that can be parsed and processed by normal execution path and at the
+        -- same time passes the information of a missing Marathon leader.
+        ngx.log(ngx.NOTICE, "Using empty Marathon leader JSON")
+        res_body = '{"leader": "not elected:0"}'
```
vespian: Either this or more code/complexity due to passing extra information to calling function :P

master/cache.lua
```diff
@@ -0,0 +1,465 @@
...
+    --      from_timer: set to true if invoked from a timer
+
+    -- Acquire lock.
+    local lock
+    if from_timer then
+        ngx.log(ngx.INFO, "Executing cache refresh triggered by timer")
+        -- Fail immediately if another worker currently holds
+        -- the lock, because a single timer-based update at any
+        -- given time suffices.
+        lock = shmlock:new("shmlocks", { timeout=0 })
```
vespian: Nice! Thanks for pointing it out!

Setting it to a constant value has some drawbacks - if for some reason, someone at some point adjusts request timeout, he/she may not be aware of `exptime` just like I did and you may not be there to catch it.

> third responds super slowly after almost timing out.

If http request lib we are using does not obey `_CONFIG.CACHE_BACKEND_REQUEST_TIMEOUT`, then unless we set it to some absurdly high value, it will never be enough.

I suggest that we set it to `3 * (_CONFIG.CACHE_BACKEND_REQUEST_TIMEOUT + 2). This way `exptime` will be always adjusted automatically. Extra 2 seconds is just an arbitrary number that is positive.

Are you OK with that?

master/cache.lua
```diff
@@ -0,0 +1,465 @@
...
+    end
+
+    if refresh_needed("marathonleader_last_refresh") then
+        fetch_and_store_marathon_leader()
+    end
+
+    local ok, err = lock:unlock()
+    if not ok then
+        -- If this fails, an unlock happens automatically,
+        -- by default after 30 seconds, to prevent deadlock.
```
vespian: will do.

vespian:
>  In this case, I don't understand why we have three levels of cache? Without this, #41 disappears entirely, why we would add this?

I think I answered it to some degree in #41:

From https://github.com/dcos/adminrouter/pull/41#issuecomment-277382059

> we need to shield our clients from temporary failures of e.g. Mesos at least in some degree. Applications do not move that often in a cluster and the cache stored for /service/<id> endpoint can be still usable for outside clients while i.e. operator tries to revive Mesos. Unifying it the way you suggest will make users using aforementioned endpoint notice the failure immediately. Having 72h CACHE_MAX_AGE_HARD_LIMIT behaviour gives some headroom to the operator. Even temporary Mesos hiccups are not so severe anymore I think.

vespian:
@gpaul @jgehrcke All done/replied, thanks for your patience!

master/cache.lua
jgehrcke: Yes!

vespian: https://github.com/dcos/adminrouter/pull/37/commits/e059b42d58035c7f4047b599dbcef71056dda381

jgehrcke:
Well, then let's start the :rocket: and see where (and how :D) it lands :tada:

vespian:
We are down from ~46 to 36 commits - I have squashed all the low-hanging fruits I could find.

Merging now.

PR #38, opened 2017-01-18
Fix exhibitor URL rewriting after introducing standardized headers.

vespian:
After introducing standardized headers the Host: field we send to exhibitor changed, and this caused proxy_redirect statement to no longer match &  fix redirects from exhibitor. This change fixes that.

The mechanism behind that is as follows:
* https://github.com/dcos/adminrouter/blob/master/common/proxy-headers.conf#L1 sets the Host header for upstream the the Host header sent by client (i.e. IP address of the cluster). This is a significant change to what we had earlier, as the name of the upstream ("exhibitor") was set.
* exhibitor assembles the "Location" header sent to client by looking on the Host header.
* https://github.com/dcos/adminrouter/blob/master/nginx.master.conf#L151 tries to match $proxy_host (in this case "exhibitor") to the redirect location sent by exhibitor (in current case it's i..e cluster IP address).
* match fails, and the unchanged "Location" header is sent  ("...exhibitor.." instead of "...exhibitor/exhibitor...")
* user feels pain


alberts:
@vespian did we find this from a dcos integration breaking? can we add a new test in your framework to catch this problem with a mock exhibitor backend?

alberts:
let's fix the mathing_header typos then we can probably :ship:

vespian:
@alberts mathing_header typo fixed.

jgehrcke:
It's a ship! :ship:

PR #41, opened 2017-01-31
DCOS-13466 Early cache population issue

vespian:
In the current setup, during the very early stage of AR start (<0,2> seconds timeframe), the cache is still empty as the first timer-triggered update has not run yet. If a request hits AR, cache update is forced to happen independently of timer-based one. This refresh is a blocking one (request waits while the cache is updated), and the timeout is set to 20 seconds - https://github.com/dcos/adminrouter/blob/master/master/mesosstatecache.lua#L233.

In theory, because of this timeout, I think it's possible to Denial-Of-Service Adminrouter in a sufficiently big cluster during this very early stage with lots of requests for i.e. `/service` endpoint. These requests will block if for example Mesos is stuck and does not respond promptly. It also does not fit our new architecture (`cache_max_age_soft_limit`, `cache_max_age_hard_limit` and friends), as during normal AR operation we reply with 500-something if Mesos decides to take it slow and requests can no longer trigger cache refresh on their own if there is a problem with Mesos/Marathon backend. So the early stage algorithm is inconsistent compared to normal error handling/there are different policies for early stage and steady stage operation.

The DOS situation can happen during both early DC/OS boot as well after AR restart.

My proposal is to try to queue requests in the early stage just like we do now, but after the first request failing, we fall back to timers and 5XX all the requests till timer-based cache refresh succeeds. This is a change compared to our current behaviour, and as such, I think it requires wider consensus.

This PR is based against https://github.com/dcos/adminrouter/pull/37, the change itself is just last two commits.

### DIFF
--- a/test-harness/tests/test_cache.py
+++ b/test-harness/tests/test_cache.py
@@ -1,6 +1,7 @@
 # Copyright (C) Mesosphere, Inc. See LICENSE file for details.
 
 import logging
+import pytest
 import requests
 import time
 
@@ -637,6 +638,60 @@ class TestCache():
 
             # Verify that the cache is OK now
             ping_mesos_agent(ar, valid_user_header)
+
+    def test_if_early_boot_stage_is_not_sensitive_to_failures(
+            self, nginx_class, valid_user_header, mocker, log_catcher):
+        # The idea here is to make Backend so slow, that AR is unable to
+        # update cache on first request. AR should then switch to fail-fast
+        # behaviour and wait for timers to update the backend when it recovers.
+
+        # This test asserts that in the case where a request is blocked,
+        # waiting for a cache update, and that update fails, then all new
+        # requests instantly fail with 503. Requests already queued will time
+        # out just the same way the first one did.
+
+        refresh_lock_timeout = 10
+        backend_request_timeout = 5
+
+        ar = nginx_class(cache_first_poll_delay=1,
+                         cache_poll_period=3,
+                         cache_expiration=2,
+                         cache_max_age_soft_limit=1200,
+                         cache_max_age_hard_limit=1800,
+                         cache_backend_request_timeout=backend_request_timeout,
+                         cache_refresh_lock_timeout=refresh_lock_timeout,
+                         )
+        agent_id = 'de1baf83-c36c-4d23-9cb0-f89f596cd6ab-S1'
+        url = ar.make_url_from_path('/agent/{}/blah/blah'.format(agent_id))
+        v = Vegeta(log_catcher, target=url, jwt=valid_user_header, rate=3)
+
+        # Make mesos slooowwwww.....
+        # It mus respond slower than backend_request_timeout
+        mocker.send_command(endpoint_id='http://127.0.0.2:5050',
+                            func_name='always_stall',
+                            aux_data=backend_request_timeout * 1.5)
+
+        with GuardedSubprocess(ar):
+            with GuardedSubprocess(v):
+                time.sleep(1)  # let it warm-up!
+
+                # Vegeta by this time should have already clogged AR
+
+                # Expect request timeout due to the cache lock holding the
+                # requests back
+                with pytest.raises(requests.exceptions.ReadTimeout):
+                    ping_mesos_agent(ar,
+                                     valid_user_header,
+                                     timeout=(10.0, backend_request_timeout * 1.5))
+
+                # Requests from now on should be failing fast
+                try:
+                    ping_mesos_agent(ar,
+                                     valid_user_header,
+                                     timeout=1,
+                                     expect_status=503)
+                except requests.exceptions.ReadTimeout:
+                    assert False, "AR does not fail-fast request after early-boot stage"
 
     def test_if_early_boot_stage_can_recover_from_a_bit_slow_backend(
             self, nginx_class, valid_user_header, mocker, log_catcher):
--- a/master/cache.lua
+++ b/master/cache.lua
@@ -431,13 +431,22 @@ function _M.get_cache_entry(name)
     -- Handle the special case of very early request - before the first
     -- timer-based cache refresh
     local entry_last_refresh = cache:get(name_last_refresh)
+
+    if entry_last_refresh == -1 then
+        ngx.log(ngx.ERR, "Aborting request due to persistent backend failure")
+        return nil
+    end
+
     if entry_last_refresh == nil then
         refresh_cache()
 
         entry_last_refresh = cache:get(name .. "_last_refresh")
         if entry_last_refresh == nil then
-            -- Something is really broken, abort!
-            ngx.log(ngx.ERR, "Could not retrieve last refresh time for `" .. name .. "` cache entry")
+            -- Backend is broken, abort!
+            if cache_data(name_last_refresh, -1) then
+                ngx.log(ngx.ERR, "Failed to limit `" .. name .. "` cache to only timer-based refresh")
+            end
+            ngx.log(ngx.ERR, "Backend seems to be broken, handling out refresh to timers")
             return nil
         end
     end


test-harness/tests/test_cache.py
```diff
@@ -637,3 +638,83 @@ def test_if_failed_request_triggered_update_is_recovered_by_timers(
...
+        # It mus respond slower than backend_request_timeout
+        mocker.send_command(endpoint_id='http://127.0.0.2:5050',
+                            func_name='always_stall',
+                            aux_data=backend_request_timeout * 1.5)
+
+        with GuardedSubprocess(ar):
+            with GuardedSubprocess(v):
+                time.sleep(1)  # let it warm-up!
+
+                # Vegeta by this time should have already clogged AR
```
gpaul: 1. Why do we need vegeta, can't we just make a known-number of requests from pure python with known behaviour?

2. What does 'clogged' mean in this case? 

3. Why does anything need to be clogged for this test to assert it's behaviour?

I would extend (the already good) test docstring as follows:

> This test asserts that in the case where a request is blocked, waiting for a cache update, and that update fails, then all queued requests instantly fail with 503.

gpaul: And if that is indeed what the test asserts, then you only need four requests in total:

The first one to trigger the cache update and block until it succeeds or fails.

The second and third to test that all queued requests fail immediately.

A fourth request which is expected to succeed N seconds later when the cache is expected to be populated (to test the case where mesos comes back after crashing.)

gpaul:
Jira link for quick reference
https://mesosphere.atlassian.net/browse/DCOS-13466

test-harness/tests/test_cache.py
vespian: Regarding your first comment:
1. There are a couple of reasons why:
  * I did not want to write new code. Subprocess class is already there, Vegeta is highly configurable. Recreating constant-rate request functionality in Python reliably would take some time.
  * We already had some discussions about GIL, Python, and single-process performance. Using Vegeta seemed future proof - no matter how much more stuff we do in the main Python process, benchmark tool is already a separate process, and we do not have to worry about timing, scheduling, etc... 
  * In this particular case we just wanted to keep the requests hanging on AR, but in future, we may want to benchmark AR components with higher requests-rate. Vegeta is in Go, and we can give AR a nice run for its money this way.
2. It's a metaphor really - imagine that you want to test if your sink is clogged by pouring water at a constant rate and checking if increasing the amount of water that goes from the tap you increase the amount of water that collects in the sink/can't enter the drainage. Here we have something similar - the constant request rate against AR creates a steady "failure state" against we issue our own, extra request and we test how AR handles it. Test this way has to only issue a single request instead of multiple in order to check if we are timing out or 5XX-ing
3. See the last sentence of the last paragraph.

Regarding your second comment:
This is true, I did not say this is the only way ;) The problem I see here is:
a) you need to pay more attention to timing if you want to get reliable test results. If you just flood AR with requests, things get easier I think.
b) you would have to somehow issue the request and then context-switch to `sleep() + issuing another request` while first one is still ongoing. I am sure there is an easy way to do it in Python (something select()-ish?), but I would have to read a bit/google it, nothing comes to my mind immediately.

I hope it helps.

vespian:
> Ideally I would like to discuss this issue and PR in Jira so we have it linked to appropriate issues and available for search. In this case, I'll copy this conversation to Jira afterwards as the related ticket lives in the DC/OS Enterprise Jira and it would benefit future maintainers of this open source project to have this, too.

Np. Thanks for the review.

test-harness/tests/test_cache.py
vespian: > This test asserts that in the case where a request is blocked, waiting for a cache update, and that update fails, then all queued requests instantly fail with 503.

kk

gpaul:
Before we discuss the solution, I would like us to discuss to problem. This is a very tricky part of a very tricky subsystem and I am not at all an expert in anything related to this PR. I have tried hard to understand the issue. Please don't take this comment as patronizing when it is intended to be clear and verbose as we can save a lot of time by communicating super clearly about this very complicated distributed systems problem. 

From the PR description I gather two problems, which I would like us to discuss separately.

The first problem is a "Denial-Of-Service (of) Adminrouter".

A DOS implies that Adminrouter (ie. nginx + a bunch of openresty/lua code) becomes unavailable from the point of view of clients making new requests. This can be because nginx fails to accept new requests entirely, due to running out of some resource (eg. memory or hitting its open file descriptor limit,) or because it fails to respond to new requests before those requests timeout for unexpected reasons, typically due to high CPU load.

From the rest of the description I gather that the problem is not that Adminrouter is overloaded in terms of CPU and as such I treat the issue that it may enter a state where new requests are not being served at all.

I see two ways in which this can happen:

One option is that nginx has crashed. For example, it could run out of memory. I don't think this is what this PR is intended to fix.

The other scenario that I can conceive of is that Adminrouter has reached its open file descriptor limit. This is very possible and I am going to focus on this issue. If this is not the issue this PR is intended to address, stop reading here and just reply saying "Woops, sorry bro, you're missing the point."

The scenario I believe we'd like to guard against is the following:

1. Adminrouter boots. It starts a separate thread (a logical thread, obviously this is not an OS thread) that polls the backend (ie. mesos and marathon) for a some data that it wishes to populate the cache with. I call this thread the 'cache refresh thread.' This thread polls the backend occasionally. The first time it runs is when the timer fires the first time. This happens after the first delay. This means that there is a number of seconds upon boot during which the cache is not populated, and no `refresh_cache()` is in process yet.

2. A request arrives. It takes the cache lock, which is uncontested at this point. It then performs the same `refresh_cache()` operation that normally occurs in the dedicated 'cache refresh thread.' For some reason, and there may be many, the backend is very slow in responding and the internal request to the backend blocks. Obviously the client request now also blocks. At this point, the cache lock is still held.

3. A second request arrives. It attempts to take the cache lock and blocks as the first request still holds the lock. It will block for a maximum of `_TIMERS.CACHE_REFRESH_LOCK_TIMEOUT`.

4. Several more requests arrive. This could be thousands or tens of thousands so let's just say an unbounded number of requests. Upon arrival, they too attempt to take the cache lock. They are all blocked trying to take the lock, as the first request still holds it.

5. After some time, the first request's internal request to the backend times out. It releases the lock and returns an appropriate HTTP error to the client.

6. The moment the first request releases the lock, all remaining requests race to acquire it (or perhaps the second request immediately acquires it, I don't know the details of the cache implementation we use but it is immaterial.)

7. At this point, whichever request newly acquires the lock request tries to refresh the cache. This operation blocks once more as the backend is still slow to respond.

8. Regardless, all requests wait a bounded amount of time, `_TIMERS.CACHE_REFRESH_LOCK_TIMEOUT` to be exact, before failing and returning appropriate HTTP errors to the client.

There is a pretty obvious DOS scenario here: the number of requests is unbounded. Every request has a local socket on the server. A socket has an accompanying open file descriptor identifying it to the operating system. This means that if requests arrive at Adminrouter at a rate that is faster than `max open files / _TIMERS.CACHE_REFRESH_LOCK_TIMEOUT` then the request queue will grow until eventually the file descriptor limit is reached and no new requests will be accepted. If Adminrouter has no limit configured then since open file descriptors are a limited system-wide shared resource other software in DC/OS will stop functioning, too.

This is obviously a really bad thing.

This PR intends to prevent this scenario by immediately failing all requests after the initial request's cache refresh fails. Requests will then "fail fast" until the 'cache refresh thread' succeeds in populating the cache. This prevents the request queue from growing after the first attempt to refresh the cache has failed.

This PR fails to guard against the case where enough requests arrive in the first period, when the first request's attempt to populate the cache has not yet succeeded, nor failed.

Instead of modifying the cache logic, I suggest that we simply configure a sensible limit (say 1,000 for purposes of this comment) to the number of requests in the nginx request queue. This will mean that the first 1,000 requests will wait in the queue where they will timeout after a while. Any further requests will immediately be refused, with the clients receiving appropriate errors.

I would now like to address the second problem laid out in the PR description:

>  It also does not fit our new architecture (cache_max_age_soft_limit, cache_max_age_hard_limit and friends), as during normal AR operation we reply with 500-something if Mesos decides to take it slow and requests can no longer trigger cache refresh on their own if there is a problem with Mesos/Marathon backend. So the early stage algorithm is inconsistent compared to normal error handling/there are different policies for early stage and steady stage operation.

I find our later stage algorithm strange - not the early stage algorithm.

My reasoning is as follows:

As client software communicating through Adminrouter I must be resilient to the 20 second lag (ie. `_TIMERS.CACHE_REFRESH_LOCK_TIMEOUT`) that may occur when Adminrouter starts up. As pointed out in the PR, this could be after an Adminrouter restart long after the system boots, when the cluster is completely stable in all other respects.

As this is the case, why do we have this artificial distinction between the early and late stages? The obvious strategy to me seems the strategy I described above.

vespian:
> This is very possible and I am going to focus on this issue. If this is not the issue this PR is intended to address, stop reading here and just reply saying "Woops, sorry bro, you're missing the point."

Yes and no - potential DOS is only a part of the problem, I do not want AR to have different strategies for dealing with cache brokerage depending on how it's early boot looked like (vide Mesos was healthy or terribly slow). I will continue reading if you do not mind ;)

> This PR intends to prevent this scenario by immediately falling all requests after the initial request's cache refresh fails. Requests will then "fail fast" until the 'cache refresh thread' succeeds in populating the cache. This prevents the request queue from growing after the first attempt to refresh the cache has failed.

jep!

> This PR fails to guard against the case where enough requests arrive in the first period when the first request's attempt to populate the cache has not yet succeeded, nor failed.

we are setting worker_connections to 1024:
https://github.com/mesosphere/adminrouter/blob/enterprise/common/main.conf#L16
https://nginx.org/en/docs/ngx_core_module.html#worker_connections

> Instead of modifying the cache logic, I suggest that we simply configure a sensible limit (say 1,000 for purposes of this comment) to the number of requests in the Nginx request queue. This will mean that the first 1,000 requests will wait in the queue where they will timeout after a while. Any further requests will immediately be refused, with the clients receiving appropriate errors.

Doable, but imagine that you have a healthy but BIG cluster (~1000 nodes). AR starts, it queues all the incoming requests while the first request waits for request-triggered cache refresh to finish. In current solution, all of them will be handled as long as we do not hit the connection limit and Mesos/Marathon responds fast enough. What you are suggesting will make some of the requests fail even though AR was able to queue them up (because their number was lower then the max-conn limit). If we set the limit to be equal to the `worker_connections` limit, then we will be just duplicating Nginx behaviour in Lua.

My idea is to limit hanging connections by time, not quantity and fall back to the long-term strategy we have.

> As client software communicating through Adminrouter I must be resilient to the 20 second lag (ie. _TIMERS.CACHE_REFRESH_LOCK_TIMEOUT) that may occur when Adminrouter starts up. As pointed out in the PR, this could be after an Adminrouter restart long after the system boots, when the cluster is completely stable in all other respects.

I think I can give you a couple of counter-arguments:
* we need to shield our clients from temporary failures of e.g. Mesos at least in some degree. Applications do not move that often in a cluster and the cache stored for `/service/<id>` endpoint can be still usable for outside clients while i.e. operator tries to revive Mesos. Unifying it the way you suggest will make users using aforementioned endpoint notice the failure immediately. Having 72h CACHE_MAX_AGE_HARD_LIMIT behaviour gives some headroom to the operator. Even temporary Mesos hiccups are not so severe anymore I think.
* imagine a long running AR in a 1000 nodes cluster that experiences a sudden Mesos borkerage. We do not control the applications that are running in the cluster and use i.e. `/service/<id>` endpoint for some external communication. A sudden spike in request response time may amplify the problem by destabilising the apps that are running in the cluster. I cannot give you concrete examples here, it's more an intuition that fail-fast on top of the shielding mentioned in the first paragraph may be a better option.

> As this is the case, why do we have this artificial distinction between the early and late stages? The obvious strategy to me seems the strategy I described above.

Totally! We just disagree which direction it should go :)

vespian:
@gpaul Thank you very much for the time you have spent reviewing it and the detailed feedback!

If you still do not feel convinced but do not want to prolong the discussion too much, we can stop here and just have a vote: me + you + @jgehrcke  form a nice quorum and if you both decide that we do not want it after reading through my answers - I am ok with that.

If you would like to continue the discussions and/or something is still unclear, the sky is the limit for me when it comes to PR comments :D 

Cheers

gpaul:
I would love to take this discussion forward, I get paid to do this afterall and even if I didn't it's just too much fun to stop now! Thanks for mentioning that we already configure the maximum number of concurrent connections.

From your response I judge that we are in agreement up to, but not including, my proposed solution. That is, the paragraph starting with "Instead of modifying the cache logic...".

Here I feel I am on rocky ground so I shall proceed with great caution.

> Doable, but imagine that you have a healthy but BIG cluster (~1000 nodes). AR starts, it queues all the incoming requests while the first request waits for request-triggered cache refresh to finish. In current solution, all of them will be handled as long as we do not hit the connection limit and Mesos/Marathon responds fast enough.

I break your response on this sentence because I was following along until this last sentence, where I stumbled. In that statement you have made some implicit assumptions which I would like to call out so we may proceed another step towards a solution. 

Not all of the requests will necessarily be handled. If the request to the backend (mesos or marathon) takes a long time, then potentially none, or maybe some, or maybe all, will be handled. The rest will timeout.

I feel responding to anything further down needs to wait for you to elaborate on what you meant here. Perhaps you could be more explicit about what requests (from clients, to the backend, etc.) timeout or get rejected in what order in your example.

vespian:
@gpaul The key to understanding my assumptions is `...Mesos/Marathon responds fast enough.` part of the quote.

> Not all of the requests will necessarily be handled. If the request to the backend (Mesos or marathon) takes a long time, then potentially none, or maybe some, or maybe all, will be handled. The rest will timeout.

This is correct and does not contradict what I said. In a `healthy BIG cluster` I think we may have a situation where there is a large thundering herd of requests hitting AR. Assuming that Mesos&Marathon are healthy, the queries for statejson, marathon apps and marathon leader will all finish in less than 20s. Provided that we do not hit the limit of connections (1024), all of them will be served. If the total time of handling all three backend requests exceeds 20s, then the thing you described will start happening - requests will start timing out. 

My point was that if we introduce the limit, you mentioned (`we simply configure a sensible limit (say 1,000 for purposes of this comment) to the number of requests in the Nginx request queue`), even though Nginx is capable of handling all the requests, we will start rejecting some of them.

I think it is worth stressing out again that queueing problems occur only during early Nginx start (restart/DC/OS boot). In "steady" state we get binary decisions - either cache is good or not and we respond accordingly.  

I hope that now it is more clear.

vespian:
> ...I judge that we are in agreement up to, but not including, my proposed solution...

I think we are, but there are a couple of factors that we need to keep in mind before arriving at the conclusions.

So, my "ideal" solution would be to make AR start handling requests only after the cache is there - no queueing, no difference in behaviour between early and stable stages, no request-triggered updates during early stages of Nginx start. Cache is updated only by timers-triggered coroutines. Client sees only binary 200/500 response to i.e. `/service/<id>` endpoint. This way we also decouple from Mesos/Marathon hiccups.

The design I described could be implemented by scheduling first update immediately in `init_by_lua` call. I guess that Nginx does not serve requests at this stage yet, but I would have to do some tests to make sure.

Also, I do not think that we have the whole context for the current code and jp could probably be helpful here, i.e. I am not sure what was the reason why we have `FIRST_POLL_PERIOD` in the first place. 

We need to keep in mind though the fact that we have limited resources. Current design works, the bug I described is a bit of a corner-case type. We should be careful to not to diverge from the original question which is "should this bugfix land?". Significant changes in cache behaviour will probably require a lot of time to discuss and properly test them.

gpaul:
> My point was that if we introduce the limit, you mentioned (we simply configure a sensible limit (say 1,000 for purposes of this comment) to the number of requests in the Nginx request queue), even though Nginx is capable of handling all the requests, we will start rejecting some of them.

Ah, I should have updated my suggestion when you mentioned the `1,024` max worker connections we already configure. This is what I suggested we do, but I didn't at the time know that we already do it. I should not be surprised :) So to be clear, I don't suggest more than one limit - the 1,024 maximum worker connections we configure right now is already what I had in mind. 

With only the current limit configured, nginx will not be capable of handling 1,025 requests, right? The last one will be refused, ie. 'fail fast'?

gpaul:
> So, my "ideal" solution would be to make AR start handling requests only after the cache is there - no queueing, no difference in behaviour between early and stable stages, no request-triggered updates during early stages of Nginx start.

I believe AR serves other requests, too. Not just to `/service/...`. These others probably shouldn't have to wait for mesos/marathon. If nginx could serve `/service/...` only once it initially populates the cache, that would work for me.

gpaul:
> I hope that now it is more clear.

Way more! Thanks for your patience, this thing is tricky. :)

gpaul:
> We need to keep in mind though the fact that we have limited resources. Current design works, the bug I described is a bit of a corner-case type. We should be careful to not to diverge from the original question which is "should this bugfix land?". Significant changes in cache behaviour will probably require a lot of time to discuss and properly test them.

Totally, I think I don't see the bug yet.

I was under the impression that the DOS you see is that we run out of open file descriptors. It is clear that this is impossible as we already configure a maximum connection limit.

Now that we know the request queue is bounded, could you state the problem we are trying to solve again?

vespian:
> With only the current limit configured, nginx will not be capable of handling 1,025 requests, right? The last one will be refused, ie. 'fail fast'?

jep, you are right. I must admit that I have not thought about it, and now I think that my DOS argument is not as strong as I thought. Though I think that we still may have reasons to apply this patch - see below.

> I believe AR serves other requests, too. Not just to /service/.... These others probably shouldn't have to wait for mesos/marathon. If nginx could serve /service/... only once it initially populates the cache, that would work for me.

I said that it is "ideal", not necessarily "practical" :) The question here is whether we want to permit partially working service. Some services may become confused by seeing that AR is up (because endpoint A is up), but they cannot perform some queries (because `/service/<id>` endpoint is still waiting for its cache refresh. I think that the answer to this question lies outside of the scope of this simple patch. 

> Totally, I think I don't see the bug yet.
> I was under the impression that the DOS you see is that we run out of open file descriptors. It is clear that this is impossible as we already configure a maximum connection limit.
> Now that we know the request queue is bounded, could you state the problem we are trying to solve again?

We still have (taken from the PRs description):
`...requests will block if for example Mesos is stuck and does not respond promptly. It also does not fit our new architecture (cache_max_age_soft_limit, cache_max_age_hard_limit and friends), as during normal AR operation we reply with 500-something if Mesos decides to take it slow and requests can no longer trigger cache refresh on their own if there is a problem with Mesos/Marathon backend. So the early stage algorithm is inconsistent compared to normal error handling/there are different policies for early stage and steady stage operation.`

On top of that - we can still have a DOS scenario: Nginx starts, requests keep coming, both for cache-related endpoints (`/service/<id>` and friends) and normal endpoints, Mesos takes it slow. Open connections pool has been exhausted (99.9% of the limit are connections that are hanging, waiting for Mesos to recover), new cache-unrelated connections get rejected most of the time - unless they get lucky. What it worse is that we can keep this state almost indefinitely as long as Mesos is down.

gpaul:
I don't have time right now to reply fully but your last failure scenario seems totally possible, yes.

vespian:
We can add a test recreating this scenario if you see value in it. Writing it should be <30 mins of work with Vegeta+some Python.

jgehrcke:
**I)** At first I would like to comment on this honorable train of thoughts:

> So, my "ideal" solution would be to make AR start handling requests only after the cache is there 

> The design I described could be implemented by scheduling first update immediately in init_by_lua call. I guess that Nginx does not serve requests at this stage yet, but I would have to do some tests to make sure.

> Also, I do not think that we have the whole context for the current code and jp could probably be helpful here, i.e. I am not sure what was the reason why we have FIRST_POLL_PERIOD in the first place.

When I researched the nginx and OpenResty integration points and architectural behavior at the time when I introduced the Mesos state cache, I was considering to build exactly this mechanism. However, to my knowledge, it is not possible to "hold back" the nginx request processing machinery from within a certain initialization hook.

My conclusion at that time was that this idea likely is architecturally incompatible with the nginx fundamentals. It actually makes sense that there is no such mechanism -- a few hopefully helpful thoughts:

* One of the pillars of the nginx architecture is that the workers operate independently.
* That "wait, I want to do something first"-like hook would require cross-worker process synchronization which does not exist by default as of the first point.
* Oh well - can't we just install said preparation hook in the _master process_ initialization? No, that's way too early in nginx' lifecyle (and if one finds an argument for why it _should_ be possible theoretically -- I think this is impossible to do with the current OpenResty architecture).

The stated problem is a problem that _requires_ cross-worker process synchronization. We manually achieve it by explicitly using the SHM-based lock.

And then, at some point I realized that the current architecture is much nicer than "making AR start handling requests only after the cache is there": of course we would like to have all sub systems operational which do not depend on the state cache being populated -- why not? :-).

So, we ended up with a periodic cache refresh ecosystem which works _entirely_ decoupled from the canonical nginx request processing system. That's actually pretty neat :-). This decoupled subsystem is fired off with an initial timer invocation, which then starts another timer, which then ... this is the concept of recursive timer usage, recommended by @agentzh: http://stackoverflow.com/a/19060625/145400

Said initial timer invocation obviously needs a (rather arbitrary) starting point -- which is how `FIRST_POLL_PERIOD` entered the picture.

By the way, quoting @agentzh:

> Using `ngx.timer` API in `init_by_lua` is not possible because `init_by_lua` runs in the Nginx master process instead of the worker processes which do the real request processing and etc.


**II)** Now, I'd like to comment on a different topic:

> On top of that - we can still have a DOS scenario: Nginx starts, requests keep coming, both for cache-related endpoints (/service/<id> and friends) and normal endpoints, Mesos takes it slow. Open connections pool has been exhausted (99.9% of the limit are connections that are hanging, waiting for Mesos to recover), new cache-unrelated connections get rejected most of the time - unless they get lucky. What it worse is that we can keep this state almost indefinitely as long as Mesos is down.

A valid scenario, indeed. I wonder, however, if it is justified to call this "DOS" (Denial of Service). The way I understand the classical meaning of DOS is that there is a set of external entities E1 which have a significant impact on service quality so that this service appears to be not fully functional from the point of view of _other_ external entities E2. In that sense, E1 is _primarily responsible_ for the service degradation.

To clarify and to confirm: here, 1024 user agents requesting `/service/<service-id>/...` could simultaneously occupy the worker process request queue as of the blocking lock acquisition (which actually represents _waiting_ for Mesos state cache population). If that is the case, a separate user agent that tries to consume a service that does _not_ depend on the state cache (example: a browser tries to fetch the static files requires for displaying the DC/OS UI) would actually be rejected by nginx.

In the scenario which you describe the service is arguably not fully functional, but the _primary_ reason for that is that its internal state is _really(!)_ bad (Mesos does not respond...), whereas the external entities can IMO barely be called "responsible". The right priority in this case of course is to work on the root cause of the failure.

Arguably, one could _optimize_ handling this failure scenario. A correct approach seems to be to _precede_ (not replace!) the blocking cross-worker lock acquisition attempt with acquisition of a **counting semaphore** of size `S` -- so that the total number of clients that is allowed to wait for semaphore acquisition is smaller than the worker-local request queue size defined by `worker_connections` (example: `S=512`, `worker_connections=1024`). OpenResty offers a worker-local counting semaphore implementation:

https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/semaphore.md

In other words, this mechanism could be used to indeed **cap** the number of user agents that simultaneously attempt to trigger the initial state cache population. I would however refrain from doing so until it is clear which kind of user-facing problem we would solve with that. At this point, for this to become a user-facing problem, it looks like there must be about 1000 cache-consuming connection attempts _per worker process_.

In summary, I acknowledge the problem, but I am not sure if it is worth the effort to address it right now -- adding the semaphore-based logic again implies the risk of adding new bugs.

**III)** Finally, I would like to dump  more food for thought about the underlying concepts, and why I believe that adding a new state to the cache contents is not a good approach:

The simple architectural idea behind the state cache is that it has precisely two states, when looked at the actual data structure representing the cache: It is either 
* not populated or
* populated (with the last known good state)

That is, the cache itself has **no notion of an error state**. This has been a deliberate design choice which we should retain -- it simplifies dealing with the cache from within new code and I believe that it helps a lot in _not_ introducing new race conditions.

Some more of my thoughts about the nginx startup phase: if the cache is not yet populated, the procedure must be _best effort_: serve as many clients as possible, as quickly as possible (this is what the simultaneously blocking cross-worker lock acquisition in the initial architecture tries to achieve: the idea is that it _immediately_ propagates a state transition to all affected user agents). Another core idea is to error out if there is no hope (this is what the lock acquisition timeout -- resulting in an error -- achieves). The outlined counting semaphore addition would _not_ mess with these paradigms, but instead try to optimize for an exceptional case.

My thinking about the "heavily outdated" state: it is not actually a first class citizen state of the cache. Because the cache itself knows nothing about this type of state. I think about this as a peripheral addition, not essential to the system -- I think about it as heuristics for detecting a _truly_ exceptional scenario, as part of the  _interpretation_ of the cache contents. I find it helpful to assume that _any_ kind of thinking derived from this must not leak into the _actual_ fundamentals of the architecture. If one follows that paradigm, it yields an almost obvious boundary condition: the fact that we desire to warn clients about heavily outdated cache contents does _not_ justify adding a third (error) state to the cache for errors occurring during the initialization phase.

Adding a third (error) state to the cache contents adds complexity. It actually _complicates_ the situation and implies a largish risk for introducing new races. An out-of-band `cache_data(name_last_refresh, -1)` as we see it in the current PR actually is a highly suspicious candidate for a new race condition :-).

gpaul:
> We still have (taken from the PRs description):
...requests will block if for example Mesos is stuck and does not respond promptly. It also does not fit our new architecture (cache_max_age_soft_limit, cache_max_age_hard_limit and friends), as during normal AR operation we reply with 500-something if Mesos decides to take it slow and requests can no longer trigger cache refresh on their own if there is a problem with Mesos/Marathon backend. So the early stage algorithm is inconsistent compared to normal error handling/there are different policies for early stage and steady stage operation.

This, as well as the remaining issue where good endpoints are starved because of the requests to `/service` backing up the queue, would be handled well by JP's suggestion. 

WDYT @vespian?

I'm uneasy about the current patch because it modifies the cache in the get path get path (`cache_data(name_last_refresh, -1)`) and only addresses the issue partially, a band-aid not a fix:

> This PR fails to guard against the case where enough requests arrive in the first period when the first request's attempt to populate the cache has not yet succeeded, nor failed.

vespian:
Let me first answer @jgehrcke comment. In a separate reply I will answer to @gpaul  one.

> However, to my knowledge, it is not possible to "hold back" the nginx request processing machinery from within a certain initialization hook.

I suspected that. Let's keep it in mind - this is an important piece of info.

> So, we ended up with a periodic cache refresh ecosystem which works entirely decoupled from the canonical nginx request processing system. That's actually pretty neat :-). This decoupled subsystem is fired off with an initial timer invocation, which then starts another timer, which then ... this is the concept of recursive timer usage, recommended by @agentzh

Nobody here questions the recursive timer usage. Kudos for researching and implementing it :)

> So, we ended up with a periodic cache refresh ecosystem which works entirely decoupled from the canonical nginx request processing system. 

Maybe I do not follow. The cache refresh *is* coupled with request processing. Cache request can be triggered by an incoming request, and failure of the said refresh can bork the said request. This PR tries to address this issue a bit.

> By the way, quoting @agentzh:
>> Using ngx.timer API in init_by_lua is not possible because init_by_lua runs in the Nginx master process instead of the worker processes which do the real request processing, etc.

I understand that the reason why you do not perform the first cache refresh in the init_worker_by_lua block directly is that cosockets are not available yet, right? This way the first poll is just us waiting for them to be available (or using request context - whatever comes first).

Anyway - I am going to prepare another PR as soon as this (and marathon-leader one) gets merged. We dug up a lot of tribal knowledge here, and it would be great if it was converted to the Adminrouter documentation at some point. I will keep you posted.

> In the scenario which you describe the service is arguably not fully functional, but the primary reason for that is that its internal state is really(!) bad (Mesos does not respond..), whereas the external entities can IMO barely be called "responsible". The right priority in this case, of course, is to work on the root cause of the failure.

It depends on how you look at it. For me, it is more about E1s (clients requesting i.e. `/service/<id>` endpoint) blocking E2s (i.e. users trying to access UI) from accessing their resource. The reason why E1s do it (Mesos is slow to respond) is irrelevant. I respect that you may disagree with that, but for me, it fits the description you provided perfectly.

> Arguably, one could optimize handling this failure scenario. A correct approach seems to be to precede (not replace!) the blocking cross-worker lock acquisition attempt with acquisition of a counting semaphore of size S -- so that the total number of clients that is allowed to wait for semaphore acquisition is smaller than the worker-local request queue size defined by worker_connections (example: S=512, worker_connections=1024). 

@gpaul already suggested something very similar. I think that some of my response I wrote then can be reused here:

> This is correct and does not contradict what I said. In a healthy BIG cluster, I think we may have a situation where there is a large thundering herd of requests hitting AR. Assuming that Mesos&Marathon are healthy, the queries for statejson, marathon apps and marathon leader will all finish in less than 20s. Provided that we do not hit the limit of connections (1024), all of them will be served. If the total time of handling all three backend requests exceeds 20s, then the thing you described will start happening - requests will start timing out.
> My point was that if we introduce the limit, you mentioned (we simply configure a sensible limit (say 1,000 for purposes of this comment) to the number of requests in the Nginx request queue), even though Nginx is capable of handling all the requests, we will start rejecting some of them.

Your approach is viable and reasonable. I am worried that we still do not address the different request handling strategies. We will still have requests for `/service/<id>` endpoints and friends timing out or being rejected immediately depending on how Adminrouter's early boot stage looked like. The DOS aspect will not be as severe as it is now (users will still be able to use i.e. UI), but you will also be crippling a bit Adminrouter's ability to swallow/queue requests in huge clusters during their boot - only 512 will get queued instead of 1024.

Let me quote myself again:

> My idea is to limit hanging connections by time, not quantity and fall back to the long-term strategy we have.

I think that your original approach of queueing the requests up to a connection limit plus my patch is a good compromise between the potential DOS scenario and crippling healthy (but BIG) cluster. I guess that customers will not complain that much if they cannot access UI for 30 seconds during severe Mesos outage.

> I would however refrain from doing so until it is clear which kind of user-facing problem we would solve with that.

Do we have to wait till some customer complaints about it? I wrote tests recreating the scenario, so we already confirmed that it is real. I believe in writing robust software and catching bugs before my users/customers find them.

> At this point, for this to become a user-facing problem, it looks like there must be about 1000 cache-consuming connection attempts per worker process.

Totally! Add on top of it that Mesos needs to fail in a very specific way - it must delay responding to requests for at least 10 seconds. I agree it is a very corner-case scenario.

What I am worried about here is that if this scenario ever occurs, it will make debugging/fixing DC/OS cluster more difficult for the field and will make our customers feel that DC/OS is unreliable/exposes heisenbug type of behaviour.

> adding the semaphore-based logic again implies the risk of adding new bugs.

That is why we have tests now :)

> That is, the cache itself has no notion of an error state. This has been a deliberate design choice which we should retain -- it simplifies dealing with the cache from within new code and I believe that it helps a lot in not introducing new race conditions.

Please remember that I am not touching the cache content itself here - I am modifying the `.._last_updated` timestamp. This approach seemed cleaner, but if you insist we can, of course, introduce another variable/shm entry that will do the same thing.

> Another core idea is to error out if there is no hope (this is what the lock acquisition timeout -- resulting in an error -- achieves). 

EXACTLY! I just want us to be consistent here. Somehow both you and @gustav keep overlooking the `different request handling strategies` issue I am raising here. In the scenario I am describing Mesos is down, the first request timed out, yet still, we keep trying and trying indefinitely. My patch is all about giving up and transitioning to the steady state/handling over Mesos pooling to recursive-timers coroutines instead of desperately grasping to the "early boot stage" behaviour.

> Because the cache itself knows nothing about this type of state. 

> the fact that we desire to warn clients about heavily outdated cache contents does not justify adding a third (error) state to the cache for errors occurring during the initialization phase.

So what are "first citizen states" of the cache for you? Could you please name them?

> An out-of-band cache_data(name_last_refresh, -1) as we see it in the current PR actually is a highly suspicious candidate for a new race condition :-).

Can you please try to come up with an example? If you think that this code could be further improved, I will be very happy to adjust it.

vespian:
> This, as well as the remaining issue where good endpoints are starved because of the requests to /service backing up the queue, would be handled well by JP's suggestion.

I have addressed this in my response to JP's comment. Please check the paragraph about handling big clusters.

> I'm uneasy about the current patch because it modifies the cache in the get path (cache_data(name_last_refresh, -1)) and only addresses the issue partially, a band-aid not a fix:

It's a compromise between complexity, the time it takes to fix properly, and the number of things we can fix. I know that current design is not bulletproof.

We can try to address this with a compromise - we split the pool of open file descriptors in two using counting semaphore approach, but at the same time, we also make Adminrouter give up trying to update cache during FIRST_POOL_PERIOD time after first request-triggered update fails.

Both request handling policy problem and potential DOS become non-issue or at least not as severe.

If you both like it and it's something you are willing to accept I can implement it, and it would definitely make me happy and allow to sleep better at night :)

What do you think?

vespian:
And again - thank you a lot for feedback on this @gpaul and @jgehrcke  :)

gpaul:
My pleasure!

> Anyway - I am going to prepare another PR as soon as this (and marathon-leader one) gets merged. We dug up a lot of tribal knowledge here, and it would be great if it was converted to the Adminrouter documentation at some point. I will keep you posted.

:tada: :D !

>> ...even though Nginx is capable of handling all the requests, we will start rejecting some of them.
> I have addressed this in my response to JP's comment. Please check the paragraph about handling big clusters.

The argument of "having Adminrouter serve less than it could" rests on an already arbitrary, tunable limit: 1,024 max worker connections. 

Even if it served perfectly it would still only serve 1,024 requests and one could argue "Why don't we set it to 2,048 as there are requests that it is dropping that it could be serving and the server has file descriptors to spare." 

So to my mind "we could do more" is not convincing. What I would find convincing is "we can't do enough." We should determine what we want "enough" to mean, and simply configure both the counting semaphore and the max worker connections accordingly.

> It's a compromise between complexity, the time it takes to fix properly, and the number of things we can fix. I know that current design is not bulletproof.

As you pointed out, this failure condition is a rare edge condition. It seems a bit unsatisfying to add a patch to fix it that still leaves a smaller but very similar rare edge condition in place. I would expect that if we are being pedantic about correctness, let's go all the way! :D If we don't have the capacity to fix this properly right now, and this isn't a critical issue where a temporary mitigation is warranted, let's think this over a bit more in this PR (we're making excellent progress, as you point out), then we can plan and finally implement a correct solution in a week or two (or whenever we find time.)

> We can try to address this with a compromise

I'm not sure a compromise is what we're looking for here. :D A compromise implies we are looking for different outcomes, whereas I'm pretty sure we're looking for the same one: AdminRouter should adhere to some contract during startup. So far the best contract I've heard is:

- As a client requesting static assets I want to download them regardless of what other Adminrouter traffic exists. This implies being resilient to DOS from slow endpoints.
- As a client requesting `/service/...` I want to get a warm (cached) result, or an error, within 20 seconds. 

The rest of your paragraph was a technical suggestion which I'd like to respond to specifically, which is why I split the first bit out. Here follows the remainder:

>  we split the pool of open file descriptors in two using counting semaphore approach, 

SGTM

> but at the same time, we also make Adminrouter give up trying to update cache during FIRST_POOL_PERIOD time after first request-triggered update fails.

What do we gain by doing this, if we also have the first bit?

gpaul:
>>An out-of-band cache_data(name_last_refresh, -1) as we see it in the current PR actually is a highly suspicious candidate for a new race condition :-).

>Can you please try to come up with an example? If you think that this code could be further improved, I will be very happy to adjust it.

Let's say there are two threads (called Leader and Follower), both are racing into `get_cache_entry` and the cache is still unpopulated. We view this flow from Leader's point of view, while commenting on what Follower is doing behind the scenes.
```
    local entry_last_refresh = cache:get(name_last_refresh)

    if entry_last_refresh == -1 then
        ngx.log(ngx.ERR, "Aborting request due to persistent backend failure")
        return nil
    end

    if entry_last_refresh == nil then
        refresh_cache()
```
`entry_last_refresh == nil` so Leader and Follower both hit `refresh_cache()` here. In `refresh_cache()` Leader takes the lock.
```
        entry_last_refresh = cache:get(name .. "_last_refresh")
```
Let's say Leader's `refresh_cache()` failed. Then `entry_last_refresh` is currently "nil"
``` 
        if entry_last_refresh == nil then
```
We enter this condition.
```
            -- Backend is broken, abort!
            if cache_data(name_last_refresh, -1) then
```
A moment before this `cache_data` command executes, Follower has by now entered `refresh_cache()`, acquired the lock, and updated the cache, so that `cache:get(name_last_refresh)` is no longer `nil`.

Bam, race condition. Leader just overwrote Follower's successful setting of the cache :(

gpaul:
The fundamental point is that you cannot write to shared state that someone else might read if you are not holding a lock, or the operation is not atomic (eg. compare-and-swap.)

vespian:
> The argument of "having Adminrouter serve less than it could" rests on an already arbitrary, tunable limit: 1,024 max worker connections.

Most likely it was - at the beginning of Adminrouter.

DC/OS is tested in many different scenarios, i.e. 1000 nodes cluster. Many of our customers have huge clusters which may (or may not) directly or indirectly rely on this limit being at least 1024 during i.e. early startup stage. Introducing a counting semaphore lowers this limit by 50%. for some of the requests. Are you sure there are no risks here?

> As you pointed out, this failure condition is a rare edge condition. It seems a bit unsatisfying to add a patch to fix it that still leaves a smaller but very similar rare edge condition in place. I would expect that if we are being pedantic about correctness, let's go all the way! :D If we don't have the capacity to fix this properly right now, and this isn't a critical issue where a temporary mitigation is warranted, let's think this over a bit more in this PR (we're making excellent progress, as you point out), then we can plan and finally implement a correct solution in a week or two (or whenever we find time.)

During my studies, we had a very nice professor how was teaching us probabilistic calculus. He once said that telecom equipment is not designed to be 100% reliable as it would make it way too costly - only pacemakers are (or at least should be :D). With every 9 that we add to our 99.x% reliability factor there is a cost.Of course we can go full-monty and make it 100% reliable. But is it really worth it to make already obscure corner case that much reliable?

As I said - this is a compromise. If we aim for a perfect solution, I do not think we will have any.

> I'm not sure a compromise is what we're looking for here. :D A compromise implies we are looking for different outcomes, whereas I'm pretty sure we're looking for the same one: AdminRouter should adhere to some contract during startup. So far the best contract I've heard is:
> * As a client requesting static assets I want to download them regardless of what other Adminrouter traffic exists. This implies being resilient to DOS from slow endpoints.
> * As a client requesting /service/... I want to get a warm (cached) result, or an error, within 20 seconds.

Let me once again remind you my main issue with current AR code:

>Somehow both you and @gustav keep overlooking the different request handling strategies issue I am raising here. In the scenario I am describing Mesos is down, the first request timed out, yet still, we keep trying and trying indefinitely. My patch is all about giving up and transitioning to the steady state/handling over Mesos pooling to recursive-timers coroutines instead of desperately grasping to the "early boot stage" behaviour.

This is confirmed by a test. So far we concentrate a lot on potential DOS scenario and not on different request handling strategies.

Could you please comment on it?

> What do we gain by doing this, if we also have the first bit?

See the paragraph above.

vespian:
> Let's say there are two threads (called Leader and Follower), both are racing into get_cache_entry and the cache is still unpopulated. We view this flow from Leader's point of view, while commenting on what Follower is doing behind the scenes.

True :) We could argue about the details, but I am more than happy to fix this bug and any other you both find provided that we decide that we want to give the idea described here a "go!".

The setting of  cache_data(name_last_refresh, -1) should most probably be moved into "refresh_cache()" call.

test-harness/tests/test_cache.py
gpaul: > I did not want to write new code. Subprocess class is already there, Vegeta is highly configurable. Recreating constant-rate request functionality in Python reliably would take some time.

My issue is not with using Vegeta in general, obviously we should use it if we're benchmarking AR. My issue with the current approach is that we're not benchmarking anything here, nor should we be clogging any kind of sink. This just obscures the failure condition that we're testing here.

For example, let's say this test succeeds after some patching, it won't tell us anything about at what point the requests started failing fast, etc. Having a statically configured number of requests (ie. max worker connections) captures the essence of the bug perfectly without resorting to statistical reasoning (the equivalent of "just add a couple of sleeps".)

We don't need to perform requests at a constant-rate. All we need is a mock mesos endpoint that blocks indefinitely, and then make 1,024 requests to it, followed by a final request that we assert fails.

gpaul:
> Somehow both you and @gustav keep overlooking the different request handling strategies issue I am raising here. In the scenario I am describing Mesos is down, the first request timed out, yet still, we keep trying and trying indefinitely.

I did comment on this:

> As this is the case, why do we have this artificial distinction between the early and late stages? The obvious strategy to me seems the strategy I described above.

> The setting of cache_data(name_last_refresh, -1) should most probably be moved into "refresh_cache()" call.

That should do it, yes.

gpaul:
>> Somehow both you and @gustav keep overlooking the different request handling strategies issue I am raising here. In the scenario I am describing Mesos is down, the first request timed out, yet still, we keep trying and trying indefinitely. My patch is all about giving up and transitioning to the steady state/handling over Mesos pooling to recursive-timers coroutines instead of desperately grasping to the "early boot stage" behaviour.

> This is confirmed by a test. So far we concentrate a lot on potential DOS scenario and not on different request handling strategies.

> Could you please comment on it? 

I commented on this in my first review comment:

> As this is the case, why do we have this artificial distinction between the early and late stages? The obvious strategy to me seems the strategy I described above.

Like you, I want to have the simplest implementation. It should have the least rules in it's contract with the client and should also be robust against the failure scenario you pointed out.

Right now, that contract appears to be:

- "As a client requesting static assets I want to download them." This should be possible regardless of what other traffic Adminrouter traffic is serving. This implies that Adminrouter must be resilient to DOS from slow endpoints.
- As a client requesting /service/... I want to get a warm (cached) result, or an error, within 20 seconds. Conversely, I am willing to tolerate a 20 second response timeout to any request I make regardless of the cluster state.

This contract holds at all times, not just when the cluster boots, as Adminrouter may be restarted at any time and clients should be able to handle that.

If we agree on that contract, then by it I reject any "fail fast" optimization in the "late stage" case. As a special case, it adds complexity to guard against a DOS that is better fixed using the semaphore suggestion.

gpaul:
> As I said - this is a compromise. If we aim for a perfect solution, I do not think we will have any.

I agree the discussion may end at a point where we have no perfect solution, but I don't agree we've reached that point yet. 

As we pointed out earlier, there's no deadline, no pressing need to modify the current behaviour. As such, we should take our time and maybe we find a solution that solves the DOS and inconsistent early/laste-stage issues completely.

gpaul:
>> The argument of "having Adminrouter serve less than it could" rests on an already arbitrary, tunable limit: 1,024 max worker connections.

> Most likely it was - at the beginning of Adminrouter.
> 
> DC/OS is tested in many different scenarios, i.e. 1000 nodes cluster. Many of our customers have huge clusters which may (or may not) directly or indirectly rely on this limit being at least 1024 during i.e. early startup stage. Introducing a counting semaphore lowers this limit by 50%. for some of the requests. Are you sure there are no risks here?

I don't know if there's a risk here. If we're worried, I suggest we double the maximum worker connections from 1,024 to 2,048, and set the semaphore value to 1,024. It's an arbitrary number that just shouldn't stray too high.

To qualify "too high" in the preceding paragraph, let's dig into file descriptors for a moment. The default system-wide file descriptor limit on linux is `size_of_ram / 1k` to a maximum of 10% of all RAM (http://serverfault.com/questions/716578/default-value-of-proc-sys-fs-file-max). So on a 1GiB host we have `100,000` file descriptors. I don't think a DC/OS master will even boot with that little memory. :)

We recently set the _default_ file descriptor limit for all our services to 16,384. Configuring Adminrouter's `worker_connections = 2,048` would be a drop in the ocean.

PR #42, opened 2017-01-31
Added description of Service Endpoints to Readme

joerg84:


README.md
```diff
@@ -9,6 +9,29 @@ within the cluster.
 ## Ports summary
 <img src="admin-router-table.png" alt="" width="100%" align="middle">
 
+## Service Endpoints
+
+Admin Router allows marathon tasks to define custom service UI and HTTP endpoints, which are made available via `<dcos-cluster>/services/<service-name>`. This can be achieved by setting the following marathon task labels:
```
jgehrcke: Capitalize Marathon?

README.md
```diff
@@ -9,6 +9,29 @@ within the cluster.
...
+    "DCOS_SERVICE_SCHEME": "http"
+  }
+```
+
+In this case `http://<dcos-cluster>/services/service-name` would be forwarded to the host running the task using the first port allocated to the task.
+
+In order for the forwarding to work reliably across task failures, we recommend co-locating the endpoints with the task. This way, if the task is restarted on a potentially other host and with different ports, Admin Router will pick up the new labels and update the routing. NOTE: Due to caching there might be an up to 30-second delay until the new routing is working.
+
+We would recommend having only a single task setting these labels for a given `service-name`.
+In the case of multiple task instances with the same `service-name` label, admin router will pick one of the tasks instances deterministically, but this might make debugging issues more difficult.
```
jgehrcke: "admin router" vs. "Admin Router" -- we should be consistent here.

jgehrcke:
Left some post merge comments. Comments on https://github.com/dcos/dcos-docs/pull/878 seem related :).

PR #46, opened 2017-02-07
Make history service endpoint use leader.mesos node as upstream

vespian:
DCOS-13598


vespian:
As it has been approved I am merging it now.

PR #48, opened 2017-02-13
Backport of some of the stuff from EE, small fixes.

vespian:
Related Jira's:
* https://jira.mesosphere.com/browse/DCOS-11921 - Add tests for Open AR
* https://jira.mesosphere.com/browse/DCOS-9269 - Merge of the Open AR refactoring patches 

Stuff included here:
* drop resty.http.simple in favour of resty.http so that one library is used by both flavours
* tests that do not test auth related things should use superuser header in order to get auth out of the picture
* agent.lua should use DEFAULT_SCHEME in order to minimize variance between EE and Open
* fix a bug with marathon upstream definition
* other, minor fixes - check the commit list.


test-harness/tests/conftest.py
```diff
@@ -246,9 +246,20 @@ def f(uid, *args, **kwargs):
 
 
 @pytest.fixture(scope='session')
+def superuser_user_header(valid_jwt_generator):
+    """This fixture further simplifies JWT handling by providing a ready-to-use
+    headers with a valid  superuser JSON Web Token for `requests` module to
+    use"""
+    token = valid_jwt_generator(uid='root')
```
jgehrcke: Where does the relation between superuser and `root` come from? That concept is not known in DC/OS. Is that a relation built into the test harness/mock system?

docker/adminrouter-upstreams.conf
```diff
@@ -7,6 +7,11 @@ proxy_ssl_verify_depth 5;
 # For the sake of tests simplicity, we will not be using/testing leader.mesos
 # alias here. This should be done by integration tests.
 set $upstream_mesos http://127.0.0.2:5050;
-set $upstream_marathon http://127.0.0.1:8443;
+set $upstream_marathon http://127.0.0.1:8080;
```
jgehrcke: So, `http://127.0.0.1:8443` was actually wrong as of mismatch between scheme and port, yes?

master/cache.lua
```diff
@@ -257,7 +260,7 @@ end
 local function fetch_and_store_state_mesos()
     -- Fetch state JSON summary from Mesos. If successful, store to SHM cache.
     -- Expected to run within lock context.
-    local mesosRes, err = request("leader.mesos", 5050, "/master/state-summary", false)
+    local mesosRes, err = request("http://leader.mesos:5050" .. "/master/state-summary", false)
```
jgehrcke: Can't we do

```
"http://leader.mesos:5050/master/state-summary"
```
instead of
```
"http://leader.mesos:5050" .. "/master/state-summary"
```
?

nginx.master.conf
```diff
@@ -171,7 +171,7 @@ http {
             # Supports chunked encoding and stream for debugging API
             proxy_request_buffering off;
 
-            proxy_pass http://$agentaddr:$agentport;
+            proxy_pass $agentaddr:$agentport;
```
jgehrcke: So, `$agendaddr` contains the URL scheme?

test-harness/tests/test_master.py
```diff
@@ -19,45 +19,45 @@ def test_if_exhibitor_endpoint_redirects_req_without_slash(
...
-        r = requests.get(url, allow_redirects=False, headers=valid_user_header)
+        r = requests.get(url, allow_redirects=False, headers=superuser_user_header)
 
         assert r.status_code == 307
         assert r.headers['Location'] == location_expected
 
     def test_if_request_is_sent_to_correct_upstream(self,
                                                     master_ar_process,
-                                                    valid_user_header):
+                                                    superuser_user_header):
```
jgehrcke: Nit: `superuser_header` instead of `superuser_user_header`?

test-harness/tests/conftest.py
vespian: This has a couple of layers. I will try to answer your questions in one go:

> I'm curious: how does auth get out of the picture with the superuser header?

This is hard to see in Open AR test harness, but EE has a Bouncer Mock that supports giving/revoking permissions (https://github.com/mesosphere/adminrouter/blob/prozlach/synchronize_with_open-tests_for_ee/test-harness/modules/mocker/endpoints/ee/iam.py). `root/superuser/admin/etc...` does not matter much for Open as all users are equal, for EE root is:
* always there, by default
* it does not care about permissions that endpoint require

Thus you can say "get out of the picture" because everything is permitted for "root", both in EE and Open.

The reason why there is no differentiation between Open and EE when it comes to user names is that some tests are used by both EE and Open (`overlay` concept) and it does not seem be woth it to use different user - Open does not care how user is named as long it's there, EE needs this user to be 'root' in order to "get out of the picture".

As for the reason why we want to `get out of the picture` at all - auth tests are done differently for EE and Open and should be tested independently anyway. So we split testing if endpoint is properly authn/authz-ing from testing if endpoint does what it is supposed to.

> Where does the relation between superuser and root come from?

This could be anything really: `admin`, `superuser`, `bootstrapuser`, `root`. The `root` came to my mind as a best choice because people writting tests are accustomed to the fact that `root` can do anything. It would be hard to explain `bootstrapuser` to outsiders, `admin` is not as strong as `root` I think.
If you insist and have good arguments we can change it, otherwise I would like to leave it as-is.

> That concept is not known in DC/OS.

Think `user which belongs to "superusers" group`.

> Is that a relation built into the test harness/mock system?

It's a relation that the mock/test-harness tries to mimick.

I hope it clarifies a bit.

docker/adminrouter-upstreams.conf
vespian: In theory yes for both, in practice yes only for `only port` part as test harness does not do https in general, only for particular tests.

master/cache.lua
vespian: I knew that you will point this out :D 

Try diffing EE and Open. Writing it this way will make it a bit easier for me to merge it with EE at some point.

nginx.master.conf
vespian: Yes, this once again prepares for merge with EE.

test-harness/tests/test_master.py
vespian: I tried to be consistent with `valid_user_header`.

What do you think about renaming it to `root_user_header`?

vespian:
> It would be great if you could update the README in https://github.com/dcos/adminrouter/tree/master/resty, using the current format, for the new HTTP library.

will do.

All done, please have another look.

PR #50, opened 2017-02-15
DNS Resolver

mhrabovcin:
Update static `leader.mesos` name in nginx configuration to variables so it will get re-resolved based on record TTL.


jgehrcke (changes requested):
Hey Martin, I quickly went through that and left a few comments. 

If everyone agrees that we go ahead with this -- then we must definitely add an expressive comment or a README section about this DNS re-resolution trick through variable usage.

nginx.master.conf
```diff
@@ -4,26 +4,18 @@ include common/main.conf;
...
-    }
-
-    upstream dcos_history_service {
-        server leader.mesos:15055;
+        server 127.0.0.1:8080;
     }
 
     upstream mesos_dns {
-        server master.mesos:8123;
+        server 127.0.0.1:8123;
```
alberts: Agreed. I think we want this change, but let's separate any changes from master.mesos to 127.0.0.1 into a separate PR

nginx.master.conf
```diff
@@ -4,26 +4,18 @@ include common/main.conf;
 http {
     # Without this, cosocket-based code in worker
     # initialization cannot resolve leader.mesos.
-    resolver 127.0.0.1:61053;
+    resolver 127.0.0.1:61053 ipv6=off;
```
mhrabovcin: We're not doing IPv6 so this will save us some queries to DNS backend that would get empty replies anyways.

nginx.master.conf
mhrabovcin: I did this to be consistent with upstream - https://github.com/mesosphere/adminrouter/blob/enterprise/nginx.master.conf#L17-L19

nginx.master.conf
```diff
@@ -4,26 +4,18 @@ include common/main.conf;
...
-        server leader.mesos:5050;
-    }
-
     upstream marathon {
-        server master.mesos:8080;
-    }
-
-    upstream dcos_history_service {
-        server leader.mesos:15055;
+        server 127.0.0.1:8080;
```
mhrabovcin: Yes and that is still true but done through variable now: https://github.com/dcos/adminrouter/pull/50/files#diff-8a636580686895dc929d495101122338R247

jgehrcke: But then we can remove this upstream definition, can't we?

mhrabovcin: I guess that diff is borked for you somehow because this PR is removing the `dcos_history_service` upstream definition.

jgehrcke: Turns out I am borked. My brain reproducibly ignored the fact that the `-    upstream dcos_history_service` line has minus and is read. I only consciously saw the change from leader.mesos to 127.0.0.1. Sorry!

nginx.master.conf
jgehrcke: OK, then I am good with having it in this PR. 

@alberts @vespian please confirm.

jgehrcke (approved):
Great work!

nginx.master.conf
jgehrcke: Although that's a change in behavior, we're probably good with it :-).

test-harness/tests/conftest.py
```diff
@@ -95,32 +95,16 @@ def syslog_mock(log_catcher):
...
+    yield s
 
-    yield (m_53, m_61053)
-
-    m_61053.stop()
-    m_53.stop()
-
-
-@pytest.fixture(scope='session')
-def navstar_ips():
```
vespian: Why do we remove these? Please remember that agent uses them: https://github.com/dcos/adminrouter/blob/master/nginx.agent.conf#L5

test-harness/tests/conftest.py
```diff
@@ -7,10 +7,10 @@
 import pytest
 
 from runner.common import (
-    DNSMock,
```
vespian: Could you also please clean up the DNSMock class and all stuff related to dnsmasq from `docker/` container?

test-harness/modules/mocker/dns.py
```diff
@@ -0,0 +1,93 @@
...
+            logger=DNSLogger("pass"),  # Don't log anything to stdout
+            )
+
+    def start(self):
+        self.server.start_thread()
+
+    def stop(self):
+        self.server.stop()
+
+    def reply_with_leader_ip(self, leader_ip):
```
vespian: It would be nice to manipulate domain TTL in the same manner, if this is required by a test. Then we can make default very high, and use low values for tests that require it. We do not have to worry about AR caching it, as we controll AR and can simply launc our own instance the same way it's done in `test_cache.py`.

test-harness/tests/test_master.py
```diff
@@ -204,3 +205,36 @@ def test_if_upstream_headers_are_correct(self,
...
+
+        # Make sure that second requeste was forwarded to new leader which is
+        # instance of ReflectingTcpIpEndpoint
+        r = requests.get(url, allow_redirects=True, headers=valid_user_header)
+        assert r.status_code == 200
+        assert r.json()["endpoint_id"] == "http://127.0.0.5:5050"
+
+        # Revert back to the original Mesos leader
+        dns_server_mock.reply_with_leader_ip('127.0.0.2')
+        sleep(2)
```
vespian: Please have a look at `test_cache.py` and how a new AR instance is started for each tests. Using this approach you will not have to introduce artificial delays/slow down the tests.

test-harness/tests/test_master.py
```diff
@@ -204,3 +205,36 @@ def test_if_upstream_headers_are_correct(self,
...
+
+        # Make sure that second requeste was forwarded to new leader which is
+        # instance of ReflectingTcpIpEndpoint
+        r = requests.get(url, allow_redirects=True, headers=valid_user_header)
+        assert r.status_code == 200
+        assert r.json()["endpoint_id"] == "http://127.0.0.5:5050"
+
+        # Revert back to the original Mesos leader
+        dns_server_mock.reply_with_leader_ip('127.0.0.2')
```
vespian: This should be done by a fixture cleanup, have a look i.e. at: https://github.com/dcos/adminrouter/blob/master/test-harness/tests/conftest.py#L63-L74

test-harness/tests/test_master.py
```diff
@@ -204,3 +205,36 @@ def test_if_upstream_headers_are_correct(self,
...
+        url = master_ar_process.make_url_from_path('/mesos/master/state-summary')
+
+        # Make sure that request was proxied to MesosEndpoint mock which returns
+        # data that looks like actual cluster state summary
+        r = requests.get(url, allow_redirects=True, headers=valid_user_header)
+        assert r.status_code == 200
+        assert r.json()["cluster"] == "prozlach-qzpz04t"
+
+        # See MockerBase::_create_common_endpoints() where we create ReflectingEndpoint
+        dns_server_mock.reply_with_leader_ip('127.0.0.5')
```
vespian: We really need this test to be launched in a separate AR instance that is GCed after the tests finishes.

If we do it using shared AR instance (long story short - all that is in the `tests_master.py` file, more info in the README.md) we are risking creating HeisenBugs/flaky tests as change of leader.mesos may be cached i.e. by lua code.

test-harness/tests/test_master.py
```diff
@@ -204,3 +205,36 @@ def test_if_upstream_headers_are_correct(self,
...
+        # The MesosLeaderDNSServer resolves leader.mesos with TTL=1s so its
+        # enough to wait for 2 seconds here
+        sleep(2)
+
+        # Make sure that second requeste was forwarded to new leader which is
+        # instance of ReflectingTcpIpEndpoint
+        r = requests.get(url, allow_redirects=True, headers=valid_user_header)
+        assert r.status_code == 200
+        assert r.json()["endpoint_id"] == "http://127.0.0.5:5050"
```
vespian: What about testing Lua code? Do we need to do it?

https://github.com/dcos/adminrouter/blob/master/master/cache.lua#L263

test-harness/tests/test_master.py
```diff
@@ -204,3 +205,36 @@ def test_if_upstream_headers_are_correct(self,
...
+        # Make a request that should be forwarded to Mesos leader instance
+        url = master_ar_process.make_url_from_path('/mesos/master/state-summary')
+
+        # Make sure that request was proxied to MesosEndpoint mock which returns
+        # data that looks like actual cluster state summary
+        r = requests.get(url, allow_redirects=True, headers=valid_user_header)
+        assert r.status_code == 200
+        assert r.json()["cluster"] == "prozlach-qzpz04t"
+
+        # See MockerBase::_create_common_endpoints() where we create ReflectingEndpoint
```
vespian: Let's skip this. Currently there are ~15 different endpoints, we will not be able to comment all tests with "which endpoints is created where" type of comments :)

vespian (changes requested):
Left some comments.

The DNS server you wrote is briliant. No matter the fate of this PR, it would be great if it was merged into the test-harness. I only wish that I knew about it earlier :)

test-harness/modules/mocker/dns.py
```diff
@@ -0,0 +1,93 @@
...
+    """MesosLeaderResolver returns responses to leader.mesos queries"""
+
+    def __init__(self, leader_ip, ttl=5, domain=DomainName('mesos')):
+        self.leader_ip = leader_ip
+        self.ttl = ttl
+        self.domain = domain
+
+    def resolve(self, request, handler):
+        # We can respond only to leader.mesos A requests
+        if (str(request.q.qname) == (self.domain.leader + ".") and
```
mhrabovcin: I've removed remaining entries because they weren't being used in any test. Is there some other reason to include these?

test-harness/tests/conftest.py
mhrabovcin: I've added back `navstar_ips` fixture and moved the dependency to the `nginx_class` fixture. One thing I don't understand though is what is being resolved with these IP addresses in case of `agent` AR instance. I'd expect some tests to fail if these were required.

mhrabovcin:
I've addressed most of the feedback in a separate commit so its easier to review changes.

* Move new tests to a separate file 
* Run AR process per test (similar to pattern in `test_cache.py`)
* Added extra tests that are exercising DNS resolution of lua code
* Improved documentation about DNS resolution
* Cleaned up `dnsmask` dependencies
* Improved clarity of DNS mock library (TTL)

test-harness/modules/mocker/dns.py
vespian: They are not used at the moment - true.

But the thing is that we want other people to write tests as well - let's make it easier for them. It costs us very little to do it, and it may save them some time if/when they need it.

vespian: https://github.com/mesosphere/dcos-enterprise/blob/master/gen_extra/dcos-config.yaml#L88

AR agent uses `master.mesos`. Let's provide all the aliases please.

test-harness/tests/conftest.py
vespian: Please remember that we have not written all the tests yet. We have ~130, I stimate we need 600-700 to have full coverage. IIRC Agent Adminrouter is almost not tested yet.

README.md
```diff
@@ -32,6 +32,34 @@ The endpoint should only use relative links for links and referenced assets such
 
 Tasks running in nested [Marathon app groups](https://mesosphere.github.io/marathon/docs/application-groups.html) will be available only using their service name (i.e, `<dcos-cluster>/service/<service-name>`) and not considering the marathon app group name (i.e., `<dcos-cluster>/service/app-group/<service-name>`).
 
+## DNS resolution
```
vespian: ❤️ 

Thanks for writting the docs!

README.md
```diff
@@ -166,13 +194,19 @@ and `.stop()` methods during the start and stop of mocker instance
 respectively. Each endpoint can be set to respond to each and every request
 with an error (`500 Internal server error`).
 
+#### DNS mock
+
+The AR requires a working DNS server that responds correctly to
+`leader.mesos` - `A` type query and resolves current Mesos leader instance.
```
vespian: We will have to update it, see my earlier comments.

test-harness/tests/conftest.py
```diff
@@ -95,23 +95,37 @@ def syslog_mock(log_catcher):
...
-    m_53.start()
+    yield s
 
-    yield (m_53, m_61053)
+    s.stop()
 
-    m_61053.stop()
-    m_53.stop()
+
+@pytest.fixture(scope='function')
```
vespian: +1

test-harness/tests/conftest.py
jgehrcke: > we have not written all the tests yet

I understand that the tests you are referring to here are supposed to cover basic behavior.

Just noting: in case of Admin Router, "full coverage" is undefined if we talk about real-world usage scenarios. "full coverage" in terms of code coverage is great, but can easily lead to a wrong sense of confidence. Concepts, formalism plus fuzzing/chaos/statistical testing with a high sampling density are required to obtain the degree of confidence that we strive for.

test-harness/tests/test_resolver.py
```diff
@@ -0,0 +1,121 @@
...
+                url, allow_redirects=True, headers=valid_user_header)
+            assert r.status_code == 200
+            assert r.json()["cluster"] == "prozlach-qzpz04t"
+
+            dns_server_mock.reply_with_leader_ip('127.0.0.5', ttl=self.SHORT_TTL)
+            # The MesosLeaderDNSServer resolves leader.mesos with TTL=1s so
+            # its enough to wait for 2 seconds here
+            time.sleep(self.SHORT_TTL * 2)
+
+            # Make sure that second request was forwarded to new leader
```
vespian: `to the new`

test-harness/tests/test_resolver.py
```diff
@@ -0,0 +1,121 @@
...
+            # which is an instance of ReflectingTcpIpEndpoint
+            r = requests.get(
+                url, allow_redirects=True, headers=valid_user_header)
+            assert r.status_code == 200
+            assert r.json()["endpoint_id"] == "http://127.0.0.5:5050"
+
+    def test_if_mesos_leader_ar_instance_is_reresolved_in_proxy_pass(
```
vespian: What is the difference between this test and the previous one? Maybe it's too early in the morning for me :(

test-harness/tests/test_resolver.py
```diff
@@ -0,0 +1,121 @@
...
+                            func_name='record_requests')
+
+        ar = nginx_class(
+            cache_poll_period=cache_poll_period,
+            cache_expiration=(cache_poll_period - 1)
+            )
+
+        with GuardedSubprocess(ar):
+            # Let the cache warm-up:
+            time.sleep(CACHE_FIRST_POLL_DELAY + 1)
```
vespian: You can also issue a request (i.e. call `ping_mesos_agent`) and trigger the refresh earlier so there is no need make the tests slower than we need.

test-harness/tests/test_resolver.py
```diff
@@ -0,0 +1,121 @@
...
+            for _ in range(3):
```
vespian: Why three times?

vespian (commented):
Some more comments for you.

I think you did a great job at testing this change outside-in. It seems I was wrong and we do can reliably test DNS re-resolving :)

test-harness/tests/conftest.py
vespian: http://stackoverflow.com/a/90021

test-harness/tests/test_resolver.py
mhrabovcin: Its testing different configuration paths, but same resolver functionality. One difference here is that previous test requests end at Mesos servers running on port `5050` while this one tests redirecting to other AR instance running on current Mesos leader on port `80`. 

I am happy to remove this test.

vespian: I see now - no, we should definitely have it - I was missing something like this when I was writing leader endpoints tests.

I would like to suggest a different approach though - instead of copy-pasting code, we can:
* add two new reflecting endpoints which will simulate extra Mesos Master processes, make them have endpoint IDs A and B.
* parametrize the test with path(`/system/v1/leader/mesos/foo/bar`, `/mesos/master/state-summary`)
* make your DNS mock server resolve `leader.mesos` as address A
* launch AR in GuardedSubprocess context
* verify that the call to path returns endpoint A
* make your DNS mock server resolve `leader.mesos` as address B
* verify that the call to path returns endpoint A

AR will not like the fact that leader.mesos is a reflector and will bork, but this is irrelevant to our test.

The ideal solution for me would be to make Mesos/Marathon endpoints have a reflecting API location, but this is nontrivial and outside of the scope of this PR.

vespian (changes requested):
Left a reply to your comment. Please check my earlier comments too. Thanks!

mhrabovcin:
Next iteration:

* Simplified leader resolution in `proxy_pass` directive test
* Updated DNS server to include all required `*.mesos` domains

test-harness/modules/mocker/common.py
```diff
@@ -53,6 +53,13 @@ def _create_common_endpoints():
         res.append(ReflectingTcpIpEndpoint(ip='127.0.0.1', port=8181))
         # mesos
         res.append(MesosEndpoint(ip='127.0.0.2', port=5050))
+        # additional mesos instances for leader DNS changes tests
+        res.append(MesosEndpoint(ip='127.0.0.3', port=5050))
+        # mesos non-leading master that runs Mesos and AR
+        res.append(ReflectingTcpIpEndpoint(ip='127.0.0.10', port=5050))
+        res.append(ReflectingTcpIpEndpoint(ip='127.0.0.10', port=80))
```
vespian: No need for these, we already have `ip='127.0.0.2', port=80` and `ip='127.0.0.3', port=80` a couple of lines below. Every time you add a new Endpoint it makes tests start&stop a bit slower so let's keep this number at minimum.

test-harness/modules/mocker/dns.py
```diff
@@ -0,0 +1,144 @@
...
+        rtype=QTYPE.A,
+        rclass=1,
+        ttl=DEFAULT_TTL,
+        rdata=A('127.0.0.1')),
+    "slave": RR(
+        rtype=QTYPE.A,
+        rclass=1,
+        ttl=DEFAULT_TTL,
+        rdata=A('127.0.0.1')),
+}
```
vespian: 👍

test-harness/tests/test_resolver.py
```diff
@@ -0,0 +1,100 @@
...
+
+class TestNginxResolver:
+
+    # For testing purposes we'll use short DNS records TTL
+    SHORT_TTL = 1
+
+    @pytest.mark.parametrize("path,dest_port",
+                             [("/system/v1/leader/mesos/foo/bar", 80),
+                              ("/mesos/master/state-summary", 5050),
+                              ])
```
vespian: +1

vespian (approved):
Just one more thing left - please re-use the endpoints we already have.

This change looks really good, I am OK with merging it. I would like us to come back to API quering at some point but till then - what we have here is definatelly better than what we have now in production.

🚂

test-harness/modules/mocker/common.py
mhrabovcin: I understand but Mesos mocked servers running on port `5050` on both IP addresses aren't `ReflectingTcpIpEndpoint` but rather `MesosEndpoint`. This means that we couldn't use parametrized test and single `generic_correct_upstream_dest_test` for both type of URLs.

README.md
```diff
@@ -32,6 +32,39 @@ The endpoint should only use relative links for links and referenced assets such
...
+DNS server used as a resolver backed resolves the name to correct leading Mesos
+master IP address.
+
+Open source version of NGINX doesn't support periodic DNS resolution of
+upstream servers. There is a workaround that AR uses to overcome this
+limitation. The workaround relies on fact that re-resolves any URL in a
+[`proxy_pass`](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass) configuration directive that comes from a
+[variable](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#set).
+NGINX resolver respects DNS response TTL flags on each record so the resolve
+results are cached for period controlled by a upstream DNS server.
```
jgehrcke: It's true that with the workaround we start respecting the TTL emitted by the DNS server, but we can also override it.

I propose a shorter version for this paragraph that
* skips the workaround explanation (instead refers to the well-written SO answer)
* documents the visible outbome

"The open source version of NGINX does not support periodic re-resolution of hostnames used in upstream definitions. In Admin Router, we use [a workaround](http://serverfault.com/a/593003/121951) to overcome this limitation, so that hostnames used in [proxy_pass](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass) configuration directives are guaranteed to be re-resolved periodically, either according to the TTL emitted by the DNS server or, if given, according to the `valid=` argument of the [resolver](http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver) configuration."


Which brings us to the question: do we want to rely on the TTL emitted by Mesos DNS or do we want to make use of `valid=`?

jgehrcke (changes requested):
I have left a few comments w.r.t. documentation / style. Looks great otherwise, we're ready to go ahead with this from my point of view.

test-harness/modules/mocker/common.py
vespian: Then let's remove the other two - they are used for marathon leader testing if I remember correctly.

Adjusting the tests should not be hard.

mhrabovcin: Quite a lot of tests are failing:

```
38 failed, 95 passed in 50.86 seconds
```

It looks like most of cache tests expect `127.0.0.2` to be returning actual mesos state response.

vespian: k, will have a look at it later.

README.md
mhrabovcin: I've changed documentation.

In our case NGINX will be respecting TTL values from `mesos-dns` server that we control and its currently set to [`60` seconds](https://github.com/mesosphere/dcos-enterprise/blob/master/gen_extra/dcos-config.yaml#L148).

mhrabovcin:
Integration tests are passing for both Open and EE version:

https://github.com/dcos/dcos/pull/1267
https://github.com/mesosphere/dcos-enterprise/pull/619

jgehrcke (approved):

gpaul (approved):
:ship:

adam-mesos:
@mhrabovcin Do we need to update/port this to the dcos/dcos repo now that adminrouter lives there? Let's close out this PR and update https://jira.mesosphere.com/browse/DCOS-5809 when that's ready.

vespian:
Superseded by:
https://github.com/dcos/dcos/pull/1747
https://github.com/mesosphere/dcos-enterprise/pull/1205

Thanks @mhrabovcin !

PR #53, opened 2017-02-24
Adjust build scripts in order to extract openresty compile stage

vespian:
Relevant issues:
* https://jira.mesosphere.com/browse/DCOS-11919?filter=-1

This PR:
* adjusts AR build scripts so that it's possible to use them in DC/OS build environment, Jenkins tests build and developers laptop. This basically means extracting OpenResty compilation stage into a separate script.
* small bugfixes and optimisations

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@ SHELL := /bin/bash
 
 DEV_PATH := /usr/local/src
 
-DCOSAR_PYLIB_LOCAL_PATH := $(CURDIR)
-DCOSAR_PYLIB_CTR_MOUNT := /usr/local/src/adminrouter/
+DCOSAR_AR_LOCAL_PATH := $(CURDIR)
+DCOSAR_AR_CTR_MOUNT := /usr/local/adminrouter/nginx/conf/
 
 BRIDGE_DEVNAME := $(shell docker network inspect -f '{{ index .Options "com.docker.network.bridge.name" }}' bridge | awk 'NF')
 BRIDGE_IP := $(shell ip a sh dev $(BRIDGE_DEVNAME) | awk '/inet / {print $$2}' | sed 's@/.*@@')
@@ -13,10 +13,12 @@ BRIDGE_IP := $(shell ip a sh dev $(BRIDGE_DEVNAME) | awk '/inet / {print $$2}' |
 # workaround for now
 # DNS_DOCKER_OPTS := --dns=8.8.8.8 --dns=8.8.4.4
 DNS_DOCKER_OPTS := --dns=$(BRIDGE_IP) --dns=8.8.8.8 --dns=8.8.4.4
-DEVKIT_COMMON_DOCKER_OPTS := --name adminrouter-devkit \
+DEVKIT_BASE_DOCKER_OPTS := --name adminrouter-devkit \
 	$(DNS_DOCKER_OPTS) \
-	-e PYTHONDONTWRITEBYTECODE=true \
-	-v $(DCOSAR_PYLIB_LOCAL_PATH):$(DCOSAR_PYLIB_CTR_MOUNT)
+	-e NUM_CORES=2 \
+	-e PYTHONDONTWRITEBYTECODE=true
+DEVKIT_DOCKER_OPTS := $(DEVKIT_BASE_DOCKER_OPTS) \
+	-v $(DCOSAR_AR_LOCAL_PATH):$(DCOSAR_AR_CTR_MOUNT)
 
 .PHONY: clean-devkit-container
 clean-devkit-container:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -32,46 +34,64 @@ clean:
...
+				/bin/bash -c "\$$OPENRESTY_COMPILE_SCRIPT" && \
+		docker commit $$(docker ps -a -q -f name=adminrouter-devkit) \
+			mesosphere/adminrouter-devkit:full && \
+		docker rm adminrouter-devkit && \
+		docker rmi -f mesosphere/adminrouter-devkit:noresty; \
 	fi
 
 .PHONY: update-devkit
 update-devkit: clean-devkit-container
+	-docker rm adminrouter-devkit
```
mhrabovcin: What is `-` command prefix doing?

Makefile
```diff
@@ -32,46 +34,64 @@ clean:
...
 		echo "+ Building devkit image"; \
-		docker rmi -f mesosphere/adminrouter-devkit:latest; \
 		docker build \
 			--rm --force-rm \
-			-t \
-			mesosphere/adminrouter-devkit:latest ./docker/ ;\
+			-t mesosphere/adminrouter-devkit:noresty \
+			-f ./docker/Dockerfile \
+				./docker/ && \
+		docker run \
```
mhrabovcin: Wouldn't be easier to have 2 separate `Dockerfile`s ?

docker/Dockerfile
```diff
@@ -1,14 +1,23 @@
...
+    VENV_DIR=/usr/local/venv \
+    AR_BIN_DIR=/usr/local/adminrouter/ \
+    VEGETA_DOWNLOAD_SHA256=2f0a69d0ae6f0bf268b7f655bd37c0104d5568d5b2bc45bbb2c405266f74e33d \
+    VEGETA_DOWNLOAD_URL=https://github.com/tsenart/vegeta/releases/download/v6.1.1/vegeta-v6.1.1-linux-amd64.tar.gz \
+    IAM_PUBKEY_FILE_PATH=/usr/local/iam.jwt-key.pub \
+    IAM_PRIVKEY_FILE_PATH=/usr/local/iam.jwt-key.priv \
+    IAM_SHARED_SECRET_FILE_PATH=/usr/local/iam.jwt-key.shared-secret
+
+# These depend on other ENV vars, so we need a separate ENV block:
+ENV OPENRESTY_DOWNLOAD_URL=https://openresty.org/download/openresty-$OPENRESTY_VERSION.tar.gz \
```
mhrabovcin: It seems that `OPENRESTY_VERSION` is a good candidate for a [build arg](https://docs.docker.com/engine/reference/builder/#/arg) as it seems that is being used only during the build and not during the runtime.

docker/Dockerfile
```diff
@@ -1,14 +1,23 @@
...
+    OPENRESTY_DIR=/usr/local/src/openresty \
+    OPENRESTY_COMPILE_SCRIPT=/usr/local/src/build-resty.sh \
+    OPENRESTY_COMPILE_OPTS="" \
+    VENV_DIR=/usr/local/venv \
+    AR_BIN_DIR=/usr/local/adminrouter/ \
+    VEGETA_DOWNLOAD_SHA256=2f0a69d0ae6f0bf268b7f655bd37c0104d5568d5b2bc45bbb2c405266f74e33d \
+    VEGETA_DOWNLOAD_URL=https://github.com/tsenart/vegeta/releases/download/v6.1.1/vegeta-v6.1.1-linux-amd64.tar.gz \
+    IAM_PUBKEY_FILE_PATH=/usr/local/iam.jwt-key.pub \
+    IAM_PRIVKEY_FILE_PATH=/usr/local/iam.jwt-key.priv \
+    IAM_SHARED_SECRET_FILE_PATH=/usr/local/iam.jwt-key.shared-secret
```
mhrabovcin: Nice! You've just compressed at least 10 image layers.

docker/Dockerfile
```diff
@@ -70,48 +80,26 @@ COPY ./requirements-tests.txt .
...
-    && make \
-    && make install \
-    && mv -v ${AR_BIN_DIR}/nginx/conf/ ${AR_BIN_DIR}/nginx/conf.orig/ \
-    && ln -s ${AR_CONF_DIR} ${AR_BIN_DIR}/nginx/conf
+    && rm openresty.tar.gz
+
+COPY build-resty.sh $OPENRESTY_COMPILE_SCRIPT
 
 # Some AR variables that mimic DC/OS:
 COPY ./iam.jwt-key.pub /usr/local/
```
mhrabovcin: If you want to save same additional layers we could replace COPY with ADD directory instruction and running a bash script that would distribute files to right location.

Makefile
vespian: http://stackoverflow.com/a/2670143

docker/Dockerfile
vespian: We are shipping OpenResty sources inside devkit container instead of "buidinfo.json/ee.buildinfo.json". This may be already confusing for some people and moving version to Makefile and to DC/OS'es 'build' script may confuse people even more ("Where is it?"), not to mention that we will have two places where it is defined and that need to be kept in sync.

docker/Dockerfile
vespian: Thx!, will try this out tomorrow (today?).

Makefile
mhrabovcin: I guess this is a response to the https://github.com/dcos/adminrouter/pull/53/files#r102913232 comment.

vespian: Yes, I am sorry - it was late, I was tired and just mixed the answers up :(

Regarding this question - in order to keep the same buildchain for both DC/OS build and local/laptop development we need to rely on tagging containers:
* during DC/OS build, the `build` script is launched inside the adminrouter-devkit container https://github.com/mesosphere/dcos-enterprise/pull/637/files#diff-e5cec6b71cde7c47f90f9173d80309f4R11. Building new containers is unavailable there.
* local development launches the same script, but with different params: https://github.com/dcos/adminrouter/pull/53/files#diff-b67911656ef5d18c4ae36cb6741b7965R49 We could build a new container here, but we would still have keep the current approach for DC/OS build.

I hope that this clears things up a bit :)

docker/Dockerfile
vespian: I just realized that using buildargs in `dcos-builder` docker image provided/build by TC's DC/OS build chain will not support it. So one more argument why I think we should rather hardcode it.

None the less, thanks for pointing it out!

docker/Dockerfile
vespian: So, there is very little difference when it comes to COPY and ADD: http://stackoverflow.com/a/24958548 But I dug a bit and it turns out we can compact COPY layers a bit: http://stackoverflow.com/a/30316600

I have adjusted the Dockerfile accordingly.

gpaul:
Specifically between `devkit` and `update-devkit`. Maybe we can have a single point of cleanup, too.

vespian:
Thanks @gpaul and @mhrabovcin for the review!

Let's 🚜  it!

PR #55, opened 2017-02-28
Add tests for Marathon svcapps caching

mhrabovcin:
Fixes [DCOS-14066](https://jira.mesosphere.com/browse/DCOS-14066). I'll move the patch to `dcos/dcos` repository once https://github.com/dcos/dcos/pull/1291 gets merged.

### DIFF
--- a/master/cache.lua
+++ b/master/cache.lua
@@ -114,7 +114,7 @@ local function fetch_and_store_marathon_apps()
        local appId = app["id"]
        local labels = app["labels"]
        if not labels then
-          ngx.log(ngx.NOTICE, "Labels not found in app '" .. appId .. "': " .. err)
+          ngx.log(ngx.NOTICE, "Labels not found in app '" .. appId .. "'")
           goto continue
        end
 
@@ -133,12 +133,14 @@ local function fetch_and_store_marathon_apps()
           goto continue
        end
 
-       -- Lua arrays default starting index is 1 not the 0 of marathon
-       local portIdx = tonumber(portIdx) + 1
+       local portIdx = tonumber(portIdx)
        if not portIdx then
           ngx.log(ngx.NOTICE, "Cannot convert port to number for app '" .. appId .. "'")
           goto continue
        end
+
+       -- Lua arrays default starting index is 1 not the 0 of marathon
+       portIdx = portIdx + 1
 
        local tasks = app["tasks"]
 
@@ -177,7 +179,7 @@ local function fetch_and_store_marathon_apps()
 
        local port = ports[portIdx]
        if not port then
-          ngx.log(ngx.NOTICE, "Cannot find port at port index '" .. portIdx .. "' for app '" .. appId .. "'")
+          ngx.log(ngx.NOTICE, "Cannot find port at Marathon port index '" .. (portIdx - 1) .. "' for app '" .. appId .. "'")
           goto continue
        end
 
--- a/test-harness/modules/mocker/endpoints/marathon.py
+++ b/test-harness/modules/mocker/endpoints/marathon.py
@@ -13,7 +13,7 @@ from mocker.endpoints.recording import (
 # pylint: disable=C0103
 log = logging.getLogger(__name__)
 
-NGINX_TASK_ALWAYSTHERE = {
+NGINX_APP_ALWAYSTHERE = {
     "id": "/nginx-alwaysthere",
     "cmd": ("cd /opt/bitnami/nginx && harpoon initialize nginx && "
             "rm -rf /opt/bitnami/nginx/html && ln -s "
@@ -169,7 +169,7 @@ NGINX_TASK_ALWAYSTHERE = {
     ]
 }
 
-NGINX_TASK_ENABLED = {
+NGINX_APP_ENABLED = {
     "id": "/nginx-enabled",
     "cmd": ("cd /opt/bitnami/nginx && harpoon initialize nginx && "
             "rm -rf /opt/bitnami/nginx/html && ln -s "
@@ -352,7 +352,10 @@ class MarathonHTTPRequestHandler(RecordingHTTPRequestHandler):
 
         with ctx.lock:
             if base_path == '/v2/apps':
-                blob = self._convert_data_to_blob(ctx.data['endpoint-content'])
+                if ctx.data["endpoint-content-is-encoded"]:
+                    blob = ctx.data['endpoint-content']
+                else:
+                    blob = self._convert_data_to_blob(ctx.data['endpoint-content'])
             elif base_path == '/v2/leader':
                 if ctx.data['leader-content'] is None:
                     msg = "Marathon leader unknown"
@@ -372,22 +375,33 @@ class MarathonEndpoint(RecordingTcpIpEndpoint):
     """An endpoint that mimics DC/OS root Marathon"""
     def __init__(self, port, ip=''):
         super().__init__(port, ip, MarathonHTTPRequestHandler)
-        self._context.data["endpoint-content"] = {"apps": [NGINX_TASK_ALWAYSTHERE, ]}
-        self._context.data["leader-content"] = {"leader": "127.0.0.2:80"}
+        self._reset()
 
     def reset(self, *_):
         """Reset the endpoint to the default/initial state."""
         with self._context.lock:
             super().reset()
-            self._context.data["endpoint-content"] = {"apps": [NGINX_TASK_ALWAYSTHERE, ]}
-            self._context.data["leader-content"] = {"leader": "127.0.0.2:80"}
+            self._reset()
 
-    def enable_nginx_task(self, *_):
-        """Change the endpoint output so that it simulates extra Nginx task
+    def enable_nginx_app(self, *_):
+        """Change the endpoint output so that it simulates extra Nginx app
            running in the cluster
         """
         with self._context.lock:
-            self._context.data["endpoint-content"]["apps"].append(NGINX_TASK_ENABLED)
+            self._context.data["endpoint-content"]["apps"].append(NGINX_APP_ENABLED)
+
+    def set_encoded_apps_response(self, content):
+        """Change the response content for apps endpoint to arbitrary data
+           and disable response JSON encoding.
+        """
+        with self._context.lock:
+            self._context.data["endpoint-content"] = content
+            self._context.data["endpoint-content-is-encoded"] = True
+
+    def set_apps_response(self, apps):
+        """Change the response content for apps endpoint"""
+        with self._context.lock:
+            self._context.data["endpoint-content"] = apps
 
     def remove_leader(self, *_):
         """Change the endpoint output so that it simulates absence of the Marathon
@@ -409,3 +423,9 @@ class MarathonEndpoint(RecordingTcpIpEndpoint):
         """
         with self._context.lock:
             self._context.data["leader-content"] = 'blah blah buh buh'
+
+    def _reset(self):
+        """Reset internal state to default values"""
+        self._context.data["endpoint-content"] = {"apps": [NGINX_APP_ALWAYSTHERE, ]}
+        self._context.data["leader-content"] = {"leader": "127.0.0.2:80"}
+        self._context.data["endpoint-content-is-encoded"] = False
--- a/test-harness/tests/test_cache.py
+++ b/test-harness/tests/test_cache.py
@@ -1,10 +1,12 @@
 # Copyright (C) Mesosphere, Inc. See LICENSE file for details.
 
+import copy
 import logging
 import requests
 import time
 
 from generic_test_code import ping_mesos_agent
+from mocker.endpoints.marathon import NGINX_APP_ALWAYSTHERE
 from mocker.endpoints.mesos import EXTRA_SLAVE_DICT
 from runner.common import CACHE_FIRST_POLL_DELAY, Vegeta
 from util import LineBufferFilter, SearchCriteria, GuardedSubprocess
@@ -161,7 +163,7 @@ class TestCache():
                          )
 
         mocker.send_command(endpoint_id='http://127.0.0.1:8080',
-                            func_name='enable_nginx_task')
+                            func_name='enable_nginx_app')
         url = ar.make_url_from_path('/service/nginx-enabled/foo/bar/')
 
         with GuardedSubprocess(ar):
@@ -214,7 +216,7 @@ class TestCache():
                          )
 
         mocker.send_command(endpoint_id='http://127.0.0.1:8080',
-                            func_name='enable_nginx_task')
+                            func_name='enable_nginx_app')
         url = ar.make_url_from_path('/service/nginx-enabled/foo/bar/')
 
         with GuardedSubprocess(ar):
@@ -373,7 +375,7 @@ class TestCache():
                             aux_data=True)
 
         mocker.send_command(endpoint_id='http://127.0.0.1:8080',
-                            func_name='enable_nginx_task')
+                            func_name='enable_nginx_app')
 
         ar = nginx_class()
         url = ar.make_url_from_path('/service/nginx-enabled/bar/baz')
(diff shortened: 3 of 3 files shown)


master/cache.lua
```diff
@@ -177,7 +179,7 @@ local function fetch_and_store_marathon_apps()
 
        local port = ports[portIdx]
        if not port then
-          ngx.log(ngx.NOTICE, "Cannot find port at port index '" .. portIdx .. "' for app '" .. appId .. "'")
+          ngx.log(ngx.NOTICE, "Cannot find port at port index '" .. (portIdx - 1) .. "' for app '" .. appId .. "'")
```
jgehrcke: Thanks for fixing that. Should we maybe add "Marathon" here  (`"Cannot find port at Marathon port index..."`) to make the intent of the log message unambiguously clear?

master/cache.lua
```diff
@@ -114,7 +114,7 @@ local function fetch_and_store_marathon_apps()
        local appId = app["id"]
        local labels = app["labels"]
        if not labels then
-          ngx.log(ngx.NOTICE, "Labels not found in app '" .. appId .. "': " .. err)
+          ngx.log(ngx.NOTICE, "Labels not found in app '" .. appId .. "'")
```
jgehrcke, line 117: Nice, this whole code section was originally created in one big bad copy pasta scenario, I believe to remember.

jgehrcke (approved):
Nice! I went through this rather quickly and have left just a [nit comment](https://github.com/dcos/adminrouter/pull/55#discussion_r103468755).

master/cache.lua
mhrabovcin: Good idea, fixed.

gpaul (approved):
:ship:

test-harness/modules/mocker/endpoints/marathon.py
```diff
@@ -389,6 +392,19 @@ def enable_nginx_task(self, *_):
         with self._context.lock:
             self._context.data["endpoint-content"]["apps"].append(NGINX_TASK_ENABLED)
 
+    def apps_response_raw(self, content):
```
vespian: `Clean code` preaches that code should tell a story. Let's keep it in mind and analyze these two functions.

`apps_response_raw` at first glance seems like a getter to me - "this function returns apps_response in raw format", but only after reading the docstring and  the code, one realizes that this is not a case - this function instructs the Marathon mock treat stored reply as binary blob/do not try to convert it to JSON string. The same thing goes for `app_response` function - again, sound like a getter, but in fact this method is setting Marathon endpoint data to certain value.

What do we want o achieve here? The story in my opinion goes  "please set the Mock's `/v2/apps` response to this-and-that. In certain cases I need you to treat the input data litteraly and do not try to parse it as JSON".

So, for me an ideal solution would be a function `set_apps_response` with two parameters:
* response to set
* how the response should be treated (`binary` vs. `data`), by default `data`

We not only get a better readability but also the type of the data becomes inherent to the data itself, it's not possible to change after it was set. This makes it much harder to people writing tests to make mistakes and easier to understand the code for people reading tests.

The aux_data can parameter that is passed to the function can be anything really - we can make it a dict with these to params as keys and the input treated as `data` if the `type` key is not present.

What do you think?

test-harness/modules/mocker/endpoints/marathon.py
```diff
@@ -372,15 +375,15 @@ class MarathonEndpoint(RecordingTcpIpEndpoint):
     """An endpoint that mimics DC/OS root Marathon"""
     def __init__(self, port, ip=''):
         super().__init__(port, ip, MarathonHTTPRequestHandler)
-        self._context.data["endpoint-content"] = {"apps": [NGINX_TASK_ALWAYSTHERE, ]}
-        self._context.data["leader-content"] = {"leader": "127.0.0.2:80"}
+        self.reset()
```
vespian: Doing it this way will result in confusing logging messages e.g. here https://github.com/dcos/adminrouter/blob/master/test-harness/modules/mocker/endpoints/generic.py#L127 - `init ~= reset`.

For me, an ideal solution would be to have a separate function (e.g. `_set_data` or something) that will be called by both `__init__` and `reset`. If you are up to a task, we can refactor it now, but I would like us to be consistent and change all the Endpoint-derived classes to follow that pattern. If you would like to keep this task/PR short - let's please do not modify it in just one place - test harness is already very complex, I fear that if we do not keep it very tidy, it will quickly become difficult to use and maintain.

test-harness/tests/test_cache.py
```diff
@@ -668,3 +668,249 @@ def test_if_early_boot_stage_can_recover_from_a_bit_slow_backend(
...
+
+    def test_upstream_wrong_json(
+            self, nginx_class, mocker, superuser_user_header):
+        filter_regexp = {
+            "Cannot decode Marathon apps JSON: ": SearchCriteria(1, True),
+        }
+
+        ar = nginx_class()
+
+        # Send wrong non-json response content
```
vespian: `Set`, not `Send`.

test-harness/tests/test_cache.py
```diff
@@ -668,3 +668,249 @@ def test_if_early_boot_stage_can_recover_from_a_bit_slow_backend(
...
+        mocker.send_command(endpoint_id='http://127.0.0.1:8080',
+                            func_name='apps_response_raw',
+                            aux_data=b"wrong response")
+
+        url = ar.make_url_from_path('/service/nginx-enabled/foo/bar/')
```
vespian: Technically this makes no difference here (we are setting borked response anyway), but still - can we please use `/nginx-alwaysthere`? The `/nginx-enabled` is reserved for task that is not enabled by default and some people reading the tests may find it confusing.

test-harness/tests/test_cache.py
```diff
@@ -668,3 +668,249 @@ def test_if_early_boot_stage_can_recover_from_a_bit_slow_backend(
...
+
+        assert lbf.all_found
+        assert lbf.extra_matches == {}
+
+    def test_app_without_labels(
+            self, nginx_class, mocker, superuser_user_header):
+        filter_regexp = {
+            "Labels not found in app '/without_labels'": SearchCriteria(1, True),
+        }
+        apps = [{"id": "/without_labels"}]
```
vespian: Here and in all other tests you are assuming certain order in which apps data is processed (`if`s to be precise) and this is `a very bad thing` ™️ .

Any refactoring that does not change the behaviour, just order in which data is processed, will make all your tests hard-fail with a confusing message, even though the functionality they are testing has not changed. What is more, we may get false negatives if something is breaking the code along the processing path, but the fact that apps data is incomplete makes the cache query fail anyway.

My strong suggestion here would be to re-use `NGINX_TASK_ALWAYSTHERE` data, which we can import and modify at will. The modifications that we are going to do should be as narrow as possible - break only one thing in the apps dict, and test the resulting behaviour.

test-harness/modules/mocker/endpoints/marathon.py
```diff
@@ -352,7 +352,10 @@ def _calculate_response(self, base_path, *_):
 
         with ctx.lock:
             if base_path == '/v2/apps':
-                blob = self._convert_data_to_blob(ctx.data['endpoint-content'])
+                if ctx.data["endpoint-content-raw"]:
```
vespian: `endpoint-content-is-raw` ?

Remember that Marathon endpoint and others will only keep getting more complex as we add tests. Lets make our lifes easier by naming stuff in easy to understand manner :)

test-harness/modules/mocker/endpoints/marathon.py
vespian: `Clean code` preaches that code should tell a story. Let's keep it in mind and analyse these two functions.

`apps_response_raw` at first glance seems like a getter to me - "this function returns apps_response in raw format", but only after reading the docstring and  the code, one realizes that this is not a case - this function instructs the Marathon mock treat stored reply as binary blob/do not try to convert it to JSON string. The same thing goes for `app_response` function - again, sound like a getter, but in fact, this method is setting Marathon endpoint data to a certain value.

What do we want to achieve here? The story, in my opinion, goes  "please set the Mock's `/v2/apps` response to this-and-that. In certain cases, I need you to treat the input data literally and do not try to parse it as JSON".

So, for me an ideal solution would be a function `set_apps_response` with two parameters:
* response to set
* how the response should be treated (`binary` vs. `data`), by default `data.`'

We not only get a better readability but also the type of the data becomes inherent to the data itself, it's not possible to change after it was set. This makes it much harder to people writing tests to make mistakes and easier to understand the code for people reading tests.

The aux_data can parameter that is passed to the function can be anything really - we can make it a dict with these to params as keys and the input treated as `data` if the `type` key is not present.

What do you think?

test-harness/tests/test_cache.py
vespian: `Set`, not `Send`.

test-harness/tests/test_cache.py
vespian: Technically, this makes no difference here (we are setting borked response anyway), but still - can we please use `/nginx-alwaysthere`? The `/nginx-enabled` is reserved for task that is not enabled by default and some people reading the tests may find it confusing.

test-harness/modules/mocker/endpoints/marathon.py
```diff
@@ -372,15 +375,15 @@ class MarathonEndpoint(RecordingTcpIpEndpoint):
     """An endpoint that mimics DC/OS root Marathon"""
     def __init__(self, port, ip=''):
         super().__init__(port, ip, MarathonHTTPRequestHandler)
-        self._context.data["endpoint-content"] = {"apps": [NGINX_TASK_ALWAYSTHERE, ]}
```
vespian: Doing it this way will result in confusing logging messages e.g. here https://github.com/dcos/adminrouter/blob/master/test-harness/modules/mocker/endpoints/generic.py#L127 - `init ~= reset`.

For me, an ideal solution would be to have a separate function (e.g. `_set_data` or something) that will be called by both `__init__` and `reset`. If you are up to a task, we can refactor it now, but I would like us to be consistent and change *all* the Endpoint-derived classes to follow that pattern. If you would like to keep this task/PR short - let's please do not modify it in just one place - test harness is already very complex, I fear that if we do not keep it very tidy, it will quickly become difficult to use and maintain.

test-harness/tests/test_cache.py
vespian: Here and in all other tests you are assuming certain order in which apps data is processed (`if`s to be precise) and this is `a very bad thing` ™️ .

Any refactoring that does not change the behaviour, just order in which data is processed, will make all your tests hard-fail with a confusing message, even though the functionality they are testing has not changed. What is more, we may get false negatives if something is breaking the code along the processing path, but the fact that apps data is incomplete makes the cache query fail anyway.

My strong suggestion here would be to re-use `NGINX_TASK_ALWAYSTHERE` data, which we can import and modify at will. The modifications that we are going to do should be as narrow as possible - break only one thing in the apps dict, and test the resulting behaviour.

test-harness/tests/test_cache.py
```diff
@@ -668,3 +668,249 @@ def test_if_early_boot_stage_can_recover_from_a_bit_slow_backend(
...
+            # Trigger cache update by issuing request:
+            resp = requests.get(url,
+                                allow_redirects=False,
+                                headers=superuser_user_header)
+            assert "cache state is invalid" in resp.content.decode('utf-8')
+            assert resp.status_code == 503
+
+            lbf.scan_log_buffer()
+
+        assert lbf.all_found
```
vespian: `lbf.extra_matches == {}` is enough.

`all_found` is a bit of a legacy thing - I am wondering if it shouldn't be removed. It was meant as a simplified way to check if all matching succeded, but now I see that this is confusing people a bit.

test-harness/tests/test_cache.py
```diff
@@ -668,3 +668,249 @@ def test_if_early_boot_stage_can_recover_from_a_bit_slow_backend(
...
+            mocker (Mocker): Mocker fixture
+            auth_header (dict): Headers that should be passed to NGINX request
+        """
+        ar = nginx_class()
+
+        mocker.send_command(endpoint_id='http://127.0.0.1:8080',
+                            func_name='apps_response',
+                            aux_data={"apps": apps})
+
+        url = ar.make_url_from_path('/service/test/foo/bar/')
```
vespian: Lets be consistent here and use `/nginx-alwaysthere` task. It makes no difference - true, but it may make easier for other people to wrap their heads around test-harness.

vespian (changes requested):
I have left some comments, please have a look.

Thanks for taking care of it!

mhrabovcin:
@vespian I've addressed all your comments.

test-harness/modules/mocker/endpoints/marathon.py
```diff
@@ -372,22 +375,33 @@ class MarathonEndpoint(RecordingTcpIpEndpoint):
...
 
     def reset(self, *_):
         """Reset the endpoint to the default/initial state."""
         with self._context.lock:
             super().reset()
-            self._context.data["endpoint-content"] = {"apps": [NGINX_TASK_ALWAYSTHERE, ]}
-            self._context.data["leader-content"] = {"leader": "127.0.0.2:80"}
+            self._reset()
 
-    def enable_nginx_task(self, *_):
```
vespian, line 385: Why app instead of task? I am OK with both, but I just wonder what is reasoning behind it.

test-harness/modules/mocker/endpoints/marathon.py
```diff
@@ -372,22 +375,33 @@ class MarathonEndpoint(RecordingTcpIpEndpoint):
...
-        """Change the endpoint output so that it simulates extra Nginx task
+    def enable_nginx_app(self, *_):
+        """Change the endpoint output so that it simulates extra Nginx app
            running in the cluster
         """
         with self._context.lock:
-            self._context.data["endpoint-content"]["apps"].append(NGINX_TASK_ENABLED)
+            self._context.data["endpoint-content"]["apps"].append(NGINX_APP_ENABLED)
+
+    def set_encoded_apps_response(self, content):
```
vespian, line 393: I like it this way even better - so explicit :)

test-harness/modules/mocker/endpoints/marathon.py
```diff
@@ -409,3 +423,9 @@ def break_leader_reply(self, *_):
         """
         with self._context.lock:
             self._context.data["leader-content"] = 'blah blah buh buh'
+
+    def _reset(self):
```
vespian, line 427: Would it be a problem for you to refactor all other endpoints this way as well - lets be consistent.

test-harness/tests/test_cache.py
```diff
@@ -668,3 +670,197 @@ def test_if_early_boot_stage_can_recover_from_a_bit_slow_backend(
...
+        filter_regexp = {
+            "Labels not found in app '{}'".format(app["id"]): SearchCriteria(1, True),
+        }
+        self.assert_filter_regexp_for_invalid_app(
+            filter_regexp, app, nginx_class, mocker, superuser_user_header)
+
+    def test_app_without_service_scheme_label(
+            self, nginx_class, mocker, superuser_user_header):
+        app = self._nginx_alwaysthere_app()
+        app["labels"].pop("DCOS_SERVICE_SCHEME", None)
```
vespian, line 722: 👍 

Exactly what I had in mind :)

vespian (approved):
Great job! I am looking forward to more PRs like this one :)

test-harness/modules/mocker/endpoints/marathon.py
mhrabovcin: Marathon works primarily with `app` concept instead of `task` so I've fixed the names.

test-harness/modules/mocker/endpoints/marathon.py
mhrabovcin: I'd rather do it in separate story, this is just a patch for not having duplicate code in a single class.

vespian: Can you please create a Jira, assign it to yourself and add me as a watcher please?

mhrabovcin: https://jira.mesosphere.com/browse/DCOS-14238

mhrabovcin:
Closing in favour of https://github.com/dcos/dcos/pull/1313

